PR #212, opened 2014-12-18
Update installing.asc

berniedurfee-ge:
At least for CentOS, there are additional Yum packages that need to be added for the build to work:  perl-devel, asciidoc, and xmlto


austerus:
I've made a similar PR as ,I've had to add a similar set of dependencies for Debian (tested on Debian 7 and the latest Mint, so I'd say it goes for Ubuntu as well). I had to install asciidoc, xmlto and docbook2x (perl-devel was installed as a dependency). I can say it works on Fedora 21 as well, but there xmlto is part of Fedora Extras which is installed by default with the Fedora Workstation DVD install that I used. 

Don't know if it works only with perl-devel and asciidoc alone, I installed docbook2x as well and it worked.

ben:
@berniedurfee-ge, is this still relevant, or should we close it out?

berniedurfee-ge:
I think this can be closed out, looks like the README mentions installing the dependencies.

PR #222, opened 2014-12-24
Change message to use imperative present tense

mikecharles:
Directly above this example commit message it's stated that one should use the imperative present tense, so I think it's best to show that right away in the following example.


ben:
@mikecharles, is this still relevant, or should we close it out?

mikecharles:
Yes, it still needs to be correct in master, so I put in a new PR (#636) since it's been so long and that was easier than trying to get this PR to work. Thanks

PR #228, opened 2015-01-01
Fixed display of exponents

StevenRoddis:
`2^80` should be `2^80^` according to http://powerman.name/doc/asciidoc


ben:
Nice. Thanks!

Issue #226, opened 2014-12-31
checkout or clone?


jon-freed:
On this page:  http://git-scm.com/book/en/v2/Getting-Started-About-Version-Control

And in this sentence:  "Every checkout is really a full backup of all the data."

Shouldn't it be using the word "clone" instead of checkout?

jon-freed:
"Clone", please.  

Yes, I am a git beginner.  And with my background -- and particularly with substantial Microsoft SharePoint usage in my history -- "check out" is more synonymous with "lock" than it is with "clone" or "copy".  So, yes, I think "clone" is more appropriate in these circumstances.

ben:
Okay, I'm down. Want to submit a pull request that fixes this?

schacon:
:+1:

PR #230, opened 2015-01-01
[about-version-control.asc] change "checkout" to "clone"

jon-freed:
"Clone" is more accurate than "checkout" for the context here. (see https://github.com/progit/progit2/issues/226)


ben:
Closes #226

PR #232, opened 2015-01-12
remove Gitosis

pismute:
Progit2 no longer talks about Gitosis so that it is little bit wierd.

### DIFF
--- a/book/04-git-server/sections/git-daemon.asc
+++ b/book/04-git-server/sections/git-daemon.asc
@@ -41,7 +41,7 @@ respawn
 ----
 
 For security reasons, it is strongly encouraged to have this daemon run as a user with read-only permissions to the repositories – you can easily do this by creating a new user 'git-ro' and running the daemon as them.
-For the sake of simplicity we'll simply run it as the same 'git' user that Gitosis is running as.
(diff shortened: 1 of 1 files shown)


ben:
Nice catch. Thanks!

book/04-git-server/sections/git-daemon.asc
```diff
@@ -41,7 +41,7 @@ respawn
 ----
 
 For security reasons, it is strongly encouraged to have this daemon run as a user with read-only permissions to the repositories – you can easily do this by creating a new user 'git-ro' and running the daemon as them.
-For the sake of simplicity we'll simply run it as the same 'git' user that Gitosis is running as.
+For the sake of simplicity we'll simply run it as the same 'git' user that `git-shell` is using as.
```
michaelgwelch: I know this is closed. I just stumbled upon this while brwosing. Is 'using as' the right phrase? 'running as' seemed better. It doesn't seem grammatical now.

ben: Woops, that slipped past me. Wanna correct it?

michaelgwelch:
Sure see #234 

Probably I should have checked contribution guidelines. Let me know if commit message needs to be improved.

Issue #237, opened 2015-01-17
-date option in the section 2.3 contains the unicode dash symbol


Vest:
Hello,

There is a minor error in the section 2.3 ([Git Basics - Viewing the Commit History](https://github.com/progit/progit2/blob/master/book/02-git-basics/sections/viewing-history.asc)). The line below

`%ad` | Author date (format respects the –date= option)
---------- | ------------

uses the "dash" symbol from the Unicode page, and a simple copy-paste operation in Windows detects it as an error. The correct syntax should be "--dash=option".

Kind regards,
Vest

peterkokot:
Hello, I've opened pull request for you as noted above if that's ok.

Vest:
Hi, the fix looks fine, but I would remove the space character between = and "option" as it is written in the help: --date=(...)

peterkokot:
Ok, thanks. Space is removed.

Vest:
Thanks, I am closing the bug.

PR #238, opened 2015-01-17
Replace dash with two hyphens for author date

peterkokot:
As noted in #237 this fixes the typo for --date option.


ben:
Oh, nice. Thanks!

PR #249, opened 2015-02-04
Fix order of `git rm` options

loganhasson:
The `filter-branch` example errors out ('Unknown option ignore-unmatch') if the `--ignore-unmatch` flag does not come before the `--cached` flag.


ben:
Oooh, nice catch!

PR #251, opened 2015-02-09
Fix SHA-1 of file.txt v2 to match diagrams

nwidger:
The SHA-1 of `file.txt` in the diagrams is `9e5e6a4` but the text refers to `9e5e64a`.  Since I'm assuming either hash is fine, I've fixed the text to match the diagram since I figure that's an easier change.


ben:
Right you are! Thanks!

PR #260, opened 2015-02-26
git merge use: --squash doesn't need --no-commit

anton-trunov:
The use of the --no-commit option together with the --squash option
is redundant since according to the manual git merge --squash doesn't
automatically make commit after merging.


ben:
Nice. Thanks!

PR #263, opened 2015-03-02
Advanced merging: use -Xignore-space-change

anton-trunov:
The explanation for -Xignore-space-change and -Xignore-all-space options
was a bit reformulated.
And for most programming languages use of the -Xignore-space-change
option while merging could be more suitable than that of
the -Xignore-all-space option because it prevents different
lexemes from concatenating together (e.g. "def hello" and "defhello").

### DIFF
--- a/book/07-git-tools/sections/advanced-merging.asc
+++ b/book/07-git-tools/sections/advanced-merging.asc
@@ -40,7 +40,7 @@ $ git commit -am 'converted hello.rb to DOS'
  1 file changed, 7 insertions(+), 7 deletions(-)
 
 $ vim hello.rb
-$ git diff -w
+$ git diff -b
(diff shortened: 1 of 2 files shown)


ben:
I like it, nice catch! Just one thing: your two commits are signed using different email addresses, and one of them doesn't map to your GitHub user. If that's of concern to you, feel free to rebase, or add the `@math.tsu.ru` address to your user.

anton-trunov:
Thanks.
What if I squashed these two commits and force-updated my automerge-with-ignore-whitespace-options branch?
This would also fix the email address issue.

ben:
That'd be perfect, and you don't have to close the pull request to do that.

You should also know that #260 has this same issue.

anton-trunov:
Done.
Fixed the email address here and for #260.

PR #250, opened 2015-02-04
Rename centralized.png in chapter 5

jnavila:
The name clashes with the chapter 1.

This fixes #217

Issue #217, opened 2014-12-20
Rakefile build uses wrong images with same filename


cdwilson:
When building the book via `bundle exec rake book:build`, image files which have the same filename are incorrectly overwritten in the `/images/` directory created by the rake task.

For example, there are two `centralized.png` images, but the `01-introduction/...` one is overwritten by the `05-distributed-git/...` one, which appears in both places in the generated PDF:

```bash
$ find . -name centralized.png
./book/01-introduction/images/centralized.png
./book/05-distributed-git/images/centralized.png
./images/centralized.png <-- this is used in both places in the generated PDF
```

As you can see in the image below, the wrong `centralized.png` is inserted in the introduction chapter.

![progit_wrong_image](https://cloud.githubusercontent.com/assets/14631/5516493/6a013e50-8856-11e4-9888-b748f22683bf.png)

schacon:
Ah, good catch. We should just rename the file - it will be easier than dealing with the build issues.

cdwilson:
I didn't check for other name conflicts, this is just the first one I noticed.  If you're going to rename the file, you'll probably want to check all the other image names as well.

dennisschagt:
I've checked all the .png files in the repository. The two centralized.png files haven't been renamed yet ([centralized-1](https://github.com/progit/progit2/blob/master/book/01-introduction/images/centralized.png) [centralized-2](https://github.com/progit/progit2/blob/master/book/05-distributed-git/images/centralized.png)).

There are some other .png files (1.png, 2.png, ..., 10.png) with duplicated names, but their copies all contain the same image anyway. All of them have copies in 4 different directories: [dir1](https://github.com/progit/progit2/tree/master/callouts), [dir2](https://github.com/progit/progit2/tree/master/book/06-github/callouts), [dir3](https://github.com/progit/progit2/tree/master/book/07-git-tools/callouts), and [dir4](https://github.com/progit/progit2/tree/master/book/B-embedding-git/callouts).

PR #264, opened 2015-03-02
Renamed centralized.png

danguafer:
Renamed images/centralized_workflow.png to images/centralized.png to fix file reference.


ben:
This was changed in #250 for a fairly good reason (see #217). 

Unfortunately that broke the website, so I opened https://github.com/git/git-scm.com/pull/505. Feel free to chime in there if you want it to be fixed. In the meantime, I think the right thing to do is let the new name stay.

danguafer:
Oh, that's right! I think I read the names completely wrong. Sorry for my mistake.

PR #272, opened 2015-03-30
Fix a typo in Chapter 6, Section 3

samford:
I saw this typo while going through the material so I figured I would fix it.  I did a quick run through all the other files but nothing else jumped out at me.


ben:
:+1: Nice catch, thanks!

PR #273, opened 2015-04-01
Tagging: Add code description with variables.

AnneTheAgile:
I got confused at first with the branch name being 'version'. So here I just added the meta-syntax to clarify the abstract form;
```
git checkout -b [branchname] [tagname]
````
is the same as the specific concrete form;
````
$ git checkout -b version2 v2.0.0
````


ben:
I like it. Thanks! :smile:

PR #282, opened 2015-04-12
sudo add sudo

pipedream:
Using the commands in the book, e.g
```
$ apt-get install git
```
Leaves an error message unfriendly to newbies and not evening mentioning sudo.
```bash
0 jan@osprey:~$apt-get install git
E: Could not open lock file /var/lib/dpkg/lock - open (13: Permission denied)
E: Unable to lock the administration directory (/var/lib/dpkg/), are you root?
100 jan@osprey:~$
```


ben:
:+1: Nice, that's better.

PR #289, opened 2015-04-21
A few it's -> its grammar fixes

robblanco:
Hi,

I spotted another rogue possessive while re-reading, so I did a quick search and found the following still lurking.

Regards,
Rob


ben:
:blush: Thanks!

PR #295, opened 2015-05-02
Example doesn't work on Windows 7.

axmbo:
Quotation marks doesn't work in the example on Git 1.9.5 on Windows 7. Only works with apostrophes.

Error:
Alexandre@MICROBOARD ~/projecs/ticgit (master)
$ git config --global alias.visual "!gitk"
sh.exe": !gitk: event not found

But:
Alexandre@MICROBOARD ~/projecs/ticgit (master)
$ git config --global alias.visual '!gitk'

Alexandre@MICROBOARD ~/projecs/ticgit (master)
$ git config alias.visual
!gitk


ben:
Nice catch, thanks!

PR #303, opened 2015-05-08
export-subst uses git log's custom formatter

jthill:
git archive uses git log's `pretty=format:` processor directly, and
strips the surrounding `$Format:` and `$` markup from the output.


ben:
This is great. Thanks! :smile:

jthill:
Yep. That got a big grin. Much appreciated. Thanks! :-)

PR #307, opened 2015-05-20
git supports relative submodules URLs

Thannoy:
Example: ../sibling.git

This is very helpful when server name changes depending on your location, or when repo is cloned using a fake hostname for any reason


ben:
Nice. Thanks!

PR #308, opened 2015-05-20
file:// repos are often easier to damage by mistake

Thannoy:
This commit is a proposal to insert as file:// cons the fact than damaging such repo is quite easy compared to other protocols.

Anyone can operate a rm -rf /* by mistake on its computer and delete mounted repository.

### DIFF
--- a/book/04-git-server/sections/protocols.asc
+++ b/book/04-git-server/sections/protocols.asc
@@ -56,9 +56,11 @@ If you and a co-worker are working on the same project and they want you to chec
 The cons of this method are that shared access is generally more difficult to set up and reach from multiple locations than basic network access.
 If you want to push from your laptop when you're at home, you have to mount the remote disk, which can be difficult and slow compared to network-based access.
 
-It's also important to mention that this isn't necessarily the fastest option if you're using a shared mount of some kind.
+It's important to mention that this isn't necessarily the fastest option if you're using a shared mount of some kind.
 A local repository is fast only if you have fast access to the data.
 A repository on NFS is often slower than the repository over SSH on the same server, allowing Git to run off local disks on each system.
+
+Finally, this protocol does not protect the repository against accidental damage. Every user has full shell access to the "remote" directory, and there is nothing preventing them from changing or removing internal Git files and corrupting the repository.
 
 ==== The HTTP Protocols
 


book/04-git-server/sections/protocols.asc
```diff
@@ -53,7 +53,7 @@ If you and a co-worker are working on the same project and they want you to chec
 
 ===== The Cons
 
-The cons of this method are that shared access is generally more difficult to set up and reach from multiple locations than basic network access.
+The cons of this method are that shared access is generally more difficult to set up and reach from multiple locations than basic network access, and it is easier to damage the repository.
```
ben: What about adding this as a paragraph below this one? Adding it as a clause here seems to interrupt the point being made, and I think there's more than just half a sentence of "con" here.

Thannoy: Yes, this can be added as a small paragraph just before the HTTP protocols next chapter. I think inserting it above would break a serie of linked paragraphs.

It can take the form below:
> Finally, this protocol does not protect the repository against mistaken damages, since it is accessible as any standard directory to anyone having access to it.

Maybe the word "also" in the meanwhile paragraph *"It's __also__ important to mention"* will have to be removed to increase global readability when we insert the new paragraph. What do you think?

ben: Sounds good! Push that change to this branch and ping me.

Thannoy:
Pull Request content is now updated.

book/04-git-server/sections/protocols.asc
```diff
@@ -56,10 +56,12 @@ If you and a co-worker are working on the same project and they want you to chec
 The cons of this method are that shared access is generally more difficult to set up and reach from multiple locations than basic network access.
 If you want to push from your laptop when you're at home, you have to mount the remote disk, which can be difficult and slow compared to network-based access.
 
-It's also important to mention that this isn't necessarily the fastest option if you're using a shared mount of some kind.
+It's important to mention that this isn't necessarily the fastest option if you're using a shared mount of some kind.
 A local repository is fast only if you have fast access to the data.
 A repository on NFS is often slower than the repository over SSH on the same server, allowing Git to run off local disks on each system.
 
+Finally, this protocol does not protect the repository against mistaken damages, since it is accessible as any standard directory to anyone having access to it.
```
ben: Yeah, we're getting a lot closer now. Just one rewording suggestion:

> Finally, this protocol does not protect the repository against accidental damage. Every user has full shell access to the "remote" directory, and there's nothing preventing them from changing or removing internal Git files and corrupting the repository.

Thannoy: sounds perfect!

I will update the PR and notify when done.

Thannoy:
Pull request content is now updated.

ben:
Perfect. Thanks for the contribution!

PR #309, opened 2015-05-25
use of gender-neutral pronoun 'they'

SkypLabs:
The gender-neutral pronoun 'they' is used many times in the text except in the following sentence :

`that's why she rebased in the first place.`


ben:
:+1: Thanks!

PR #310, opened 2015-05-25
fix bad subject

SkypLabs:
`Now, someone else does more work that includes a merge, and pushes that work to the central server.`

In the previous sentence, we talk about the work done by another developer, not about their implicit commits.


ben:
Nice catch. Thanks!

PR #311, opened 2015-05-27
Added examples how to change the directory

haripetrov:
It's very useful to have example for each OS how to change the directory
in Git (especially for Windows because the syntax is different from the
default syntax — C:/Users/your_repository).


book/02-git-basics/sections/getting-a-repository.asc
```diff
@@ -7,7 +7,25 @@ The second clones an existing Git repository from another server.
 
 ==== Initializing a Repository in an Existing Directory
 
-If you're starting to track an existing project in Git, you need to go to the project's directory and type
+If you're starting to track an existing project in Git, you need to go to the project's directory
```
ben: What if we lead into the examples in a bit friendlier of a way? Something like this:

> […] you need to go to the project's directory. If you've never done this, it looks a little different depending on which system you're running.

haripetrov: I totally agree with you. Your way of explanation is much better than what I did (just to add the examples). What do you prefer:
1. Me to change the pull request with additional commit?
2. You to create an additional commit after you merge the PR?
3. Anything else?

ben: If actually prefer if you made the change. That way you name is on all of
the commits for this contribution, and your name shows up in the blame
view. It's your improvement, you should get all the credit.

haripetrov:
After several mistakes with git commands I managed to delete the wrong commits and to create only one additional commit.
Nice to see how Git is working.
@ben I think that this PR is ready.

madkite:
> especially for Windows because the syntax is different from the
default syntax — C:/Users/your_repository

But in your PR:

>$ cd /c/user/your_repository
(without "s")

Also as I remember user's home dir for this OS starting from v.7 is `C:/Users/<user>/`. Maybe it's better to use %UserProfile% variable - it works in older versions:
```
$ cd $USERPROFILE/<your_repository>
```

ben:
Yeah, I hate to get too deep into OS specifics. E.g. what if the user is running Powershell? The paths will be different then, too.

@haripetrov, would you mind cleaning up the merge conflicts real quick? I'd love to merge this in.

haripetrov:
Hm, now we don't have conflicts but I am not sure if this was the right way to do it. :)

haripetrov:
I will close this off and then a new request will be created with one more change from the original one (one year ago).

PR #317, opened 2015-06-18
Should it be three instead of ones?

spacewander:


ben:
Looks good to me. Thanks!

PR #318, opened 2015-06-19
Adjust Position of Semicolon

oldsharp:
"\`\`refs;''" should be "\`\`refs'';".


ben:
Wow, [nice catch](http://blog.apastyle.org/apastyle/2011/08/punctuating-around-quotation-marks.html)!

PR #319, opened 2015-06-19
Adjust Line Break

oldsharp:
This is a minor improvement -- if you'd prefer to leave this alone, I'll close this Pull Request.


oldsharp:
This seem like to be a workaround for avoiding two "``" appear in the same line?

ben:
Possibly, but any sane renderer should be able to handle that. I guess we'll find out!

PR #323, opened 2015-06-26
fixed number error in transfer-protocols example

zwpaper:
I found the best way to fix the error is correcting the numbers, so I did it.

### DIFF
--- a/book/10-git-internals/sections/transfer-protocols.asc
+++ b/book/10-git-internals/sections/transfer-protocols.asc
@@ -144,10 +144,9 @@ It tries to run a command on the remote server via an SSH call that looks someth
 [source,console]
 ----
 $ ssh -x git@server "git-receive-pack 'simplegit-progit.git'"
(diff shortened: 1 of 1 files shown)


zwpaper:
I delete a space and a line, I made a commit each line, review it please.

ben:
Apart from including the `□` character, this looks great!

zwpaper:
added a □ to indicate the '0x00', and happened to find another two mistaken. review it please.

BTW, is the order of '00009done' and '0000' in line 251 and 252 right?
In my opinion, the '0000' should be the end of a message, but here it shows up before the '0009done'.
I have try to catch the package, but the package is encrypted.
and in line 280, there is no '0009done', should it be like this?

excuse me for that I find nothing for the answer to this question, so that I have to ask here.

zwpaper:
@ben I have find the answer in the source code of git, it should be 
```
0009done
0000
```
I will update the PR, and the sample should be finally correct!

ben:
Beautiful. Thanks!

PR #324, opened 2015-06-27
Update basic-branching-and-merging.asc

mosdalsvsocld:
The wording "merge in iss53" was ambiguous and could be read with the opposite meaning: that you should be "in iss53" when you "merge". In other systems merging might actually be done in this opposite direction, so it was very confusing.


ben:
I like it! :+1:

PR #326, opened 2015-07-06
Change [] to <> in 'git remote add' command

networm:


ben:
Hmm. We don't have a clear convention around this; there are lots of `command [arg]` forms, as well as lots of `command <arg>` forms. Is there a specific reason you want to change this one in particular?

networm:
`command [arg]` means that arg is optional, but `command <arg>` means that arg is mandatory.

If we don't have a clear convetion around this then we don't need to fix this one.

OlingCat:
@ben Look at some helps like `git help push`:

>The ssh and git protocols additionally support ~username expansion:
       ·   `ssh://[user@]host.xz[:port]/~[user]/path/to/repo.git/`

>If you want to rewrite URLs for push only, you can create a configuration section of the form:
```
[url "<actual url base>"]
        pushInsteadOf = <other url base>
```

is that a Implicit convention?

ben:
Sort of. Each example seems to have its own syntax (one uses angle brackets, the other not), but each one is fairly readable, which is I think what we were going for with Pro Git. We definitely didn't go through every code block and check for consistency.

My reasoning here is: why this one instance? Why change it *this* way, and not the other? Would this change make the book materially better, or is it just an opinion? These are the things that go through my mind when considering generous contributions from the community.

PR #329, opened 2015-07-14
New line should separate sentences (02-git-basics/sections/recording-changes.asc)

noisy:
There are few places in `book/02-git-basics/sections/recording-changes.asc` where rule mentioned in title of this PR is not fallowed. This PR fixes that, because "one sentence per line" is very helpful for translators :)


ben:
:+1: Nice catch!

PR #330, opened 2015-07-14
Bring text and example in line in undoing.asc

rmzelle:
Text uses "git add *", whereas the following example uses "git add .".

I chose "git add *" for both, but it looks like either is correct. See also http://stackoverflow.com/questions/26042390/git-add-asterisk-vs-git-add-period and http://stackoverflow.com/questions/16969768/what-does-git-add-git-add-single-dot-command-do.


ben:
Awesome, thanks!

PR #333, opened 2015-07-21
Small confusing moments

madkite:
1. We add 'test line' without a sharp sign and following git diff shows it with the sharp sign. It should be consistent.
2. Ignore rule doc/**/*.txt overrides doc/*.txt, so let's change to something clear.


ben:
Lovely. Thanks! :smile:

PR #334, opened 2015-07-22
Prefer open-source tools over proprietary software in examples

madkite:


schacon:
The point of this section is to have people recognize what the software helps to do, so having more popular options that people are more likely to identify with would be best. I think Araxis is a bit more popular than the other options so I would like to keep it, but I'm fine adding kdiff3 if you wish.

madkite:
I asked almost all my friends - nobody uses Araxis and doesn't even hear about it. In this case maybe it's better to leave only two most popular tools - emerge & vimdiff. In any way command with full list of supported tools are provided. Otherwise it looks like ad.

ben:
Here's the thing:

```
$ git difftool --tool-help
'git difftool --tool=<tool>' may be set to one of the following:
		araxis
		emerge
		opendiff
		p4merge
		vimdiff
		vimdiff2
		vimdiff3

The following tools are valid, but not currently available:
		bc
		bc3
		codecompare
		deltawalker
		diffmerge
		diffuse
		ecmerge
		gvimdiff
		gvimdiff2
		gvimdiff3
		kdiff3
		kompare
		meld
		tkdiff
		xxdiff

Some of the tools listed above only work in a windowed
environment. If run in a terminal-only session, they will fail.
```

The reason Araxis is mentioned here is because Git ships with support for it, along with a couple other commercial tools.

Okay, first principles. The goal of this passage is to show that Git works with a variety of tools, both OSS and proprietary. How do we best achieve that? What about this?

> If you run `git difftool` instead of `git diff`, you can view any of these diffs in software like  emerge, vimdiff, Araxis Merge, Beyond Compare, and many more.

We show two examples each of OSS and commercial tools, and the OSS tools are right up front. Does that satisfy all our goals?

madkite:
Git supports two dozen tools and Araxis isn't good example, because it's proprietary and it isn't very popular. For instance, meld is open source and more popular - it's most downloaded diff tool from Ubuntu repo (http://popcon.ubuntu.com). I don't think Araxis has the same number of users (licenses). If you want to point to support of commercial software let's say this without any ad.
> If you run git difftool instead of git diff, you can view any of these diffs in software like emerge, vimdiff and many more (including commercial products).

ben:
I'd be okay with that change. Push it here and we can merge it.

madkite:
Done.

PR #337, opened 2015-07-27
There is no contradiction, so no "but".

mvdw:
There is no contradiction, so no "but".
Also, the most past tense is "has been", then "was", then "is".


ben:
I like it. Thanks!

PR #346, opened 2015-08-04
Delete pdf.xsl~

rmzelle:
Pretty sure this is a temp file that was accidentally committed.

(compare https://github.com/progit/progit2/blob/master/theme/pdf/pdf.xsl and https://github.com/progit/progit2/blob/master/theme/pdf/pdf.xsl~)


ben:
There's just one line of diff between the `pdf.xsl~` and `pdf.xsl`:

```diff
44c44
<         <xsl:with-param name="by" select="'../../callouts'"/>
---
>         <xsl:with-param name="by" select="'../callouts'"/>
```

And since it looks like the PDF callouts are working just fine, this seems safe. Thanks!

PR #350, opened 2015-08-06
Add missing double dashes

nobozo:


madkite:
Why do you think it is missing? "--" is only needed to prevent misunderstanding between path and tree-ish if you omit the last one. In this case you have parameter "HEAD" before, so all other parameters are treated as paths (even if some filename will be the same as some tree-ish). Syntax allows you to always use "--", but there is no sense in it if you have already pass tree-ish.

nobozo:
Because this line is trying to show that the two highlighted commands are equivalent. Since the alias has the two dashes, then it makes sense to show them in the unaliased command. Whether or not they're needed in the first place is a different questions.

madkite:
You are right, I haven't noticed that alias contains dashes. Maybe it's better to remove these dashes from alias or add some explanation, because this is first time using dashes in "git reset" command from the beginning of the book. Or even remove "HEAD" (with leaving dashed) in order to show that this is default.

nobozo:
Since 'git reset' as a command hasn't really been discussed at all, I don't think it's relevant to explain why dashes are or aren't necessary. The text was just attempting to show how an alias relates to the unaliased equivalent.

ben:
Yeah, it's probably confusing to see the dashes above and not below. And @nobozo is right, we don't need to explain the intricacies of `git reset` here, but if someone happens upon this and copy-pastes it into their config, we want them to end up with the right alias, so the dashes are appropriate.

PR #353, opened 2015-08-09
Match text to text in image.

nobozo:
The .png file says "The initial commit of my project". This change makes the text match.


ben:
Nice catch!

PR #362, opened 2015-08-16
fix a typo: .htaccess instead of .htpasswd

iprok:
It obviously should be .htpasswd not .htaccess here


ben:
Nice catch, thanks!

PR #363, opened 2015-08-16
Correct "git filter-branch" command line argument on PowerShell

satob:
With Git for Windows (https://msysgit.github.io/), folloing command

> PS C:\tmp\src\git\pullreq> git filter-branch -f --msg-filter 'sed "s/^git-tfs-id:.*$//g"' -- --all

gets following error message:

> usage: git core\git-filter-branch [--env-filter <command>] [--tree-filter <command>]
>  ...

Command line argument '--' should be quoted like:

> PS C:\tmp\src\git\pullreq> git filter-branch -f --msg-filter 'sed "s/^git-tfs-id:.*$//g"' '--' --all

and output message will be like this:

> Rewrite c9cea454ab929f2db984184647b9da9cc581abde (8/8)
> ...

### DIFF
--- a/book/09-git-and-other-scms/sections/import-tfs.asc
+++ b/book/09-git-and-other-scms/sections/import-tfs.asc
@@ -52,7 +52,7 @@ The following command will do that:
 
 [source,powershell]
 ----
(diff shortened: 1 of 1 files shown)


ben:
Nice catch!

PR #364, opened 2015-08-17
Replaced the usage of working directory with working tree in the basi…

vogella:
According to the official Git documentation the term working tree should be used. If this PR gets merged, I'm happy to provide update to the other chapters

See http://git-scm.com/docs/git-read-tree

### DIFF
--- a/book/01-introduction/sections/basics.asc
+++ b/book/01-introduction/sections/basics.asc
@@ -82,15 +82,15 @@ Committed means that the data is safely stored in your local database.
 Modified means that you have changed the file but have not committed it to your database yet.
 Staged means that you have marked a modified file in its current version to go into your next commit snapshot.
 
-This leads us to the three main sections of a Git project: the Git directory, the working directory, and the staging area.
+This leads us to the three main sections of a Git project: the Git directory, the working tree, and the staging area.
 
-.Working directory, staging area, and Git directory.
-image::images/areas.png["Working directory, staging area, and Git directory."]
+.Working tree, staging area, and Git directory.
+image::images/areas.png["Working tree, staging area, and Git directory."]
 
 The Git directory is where Git stores the metadata and object database for your project.
 This is the most important part of Git, and it is what is copied when you clone a repository from another computer.
 
-The working directory is a single checkout of one version of the project.
+The working tree is a single checkout of one version of the project.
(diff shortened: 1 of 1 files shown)


nobozo:
I did a quick scan of http://git-scm.com/docs/git-read-tree and I don't see anywhere that says "working tree" should be used instead of "working directory". In fact, that page uses both terms! This looks like an example of Git's inconsistent terminology.

Given all this, plus the amount of work it would take to make this change throughout the whole book, I don't think this change is worthwhile doing.

vogella:
@nobozo maybe a better example is http://git-scm.com/docs/git-clone which only speaks about the working tree. Directory is also used but only to point to the current directory.

I personally think the distinction of "working tree" as checkout of the files and "working directory" as the current working directory is important for the reader. I plan to provide more patches for the other chapters, if this one gets accepted.

nobozo:
I see what you mean, although in the git-clone page "working tree" only appears two times.

However, the more I think about it, the more I like "working tree" because "working directory" implies a single level structure. "Working directory tree" would be even better but I'm not sure anybody would go for that.

In my opinion, the goal for this book, and any other book about Git, should be to use the generally accepted terminology. So, I would recommend getting the opinion of some of the big names in Git, and then using whatever they think is best.

vogella:
@nobozo I send a patch for the Documentation: read-tree to the Git mailing list. I update this PR with the result of their discussion / bashing. :-)

vogella:
@nobozo Also, when I published my Git book based on my online tutorials, +Matthias Sohn project lead for the Eclipse Git project (EGit) and JGit committer project also gave me the feedback that working directory is very confusing for the reader, exact because it implies a flat structure, like you also said.

nobozo:
You've totally convinced me. I'm going to make this change in the "Pro Git Reedited 2nd Ed." book I'm working on (if I ever finish it).

vogella:
Great news. Do you want to use this PR, or shall I abondan it?

nobozo:
I think you should keep it. The book I'm working on is a separate fork of this.

vogella:
FYI My use "working tree instead of working directory" patch got merged into Eclipse Git. http://git.eclipse.org/c/egit/egit.git/commit/?id=e71a75a6311770a6ca14471416832381c61f90f1 … Making the world a better place.

ben:
OK, I'm on board.

PR #370, opened 2015-08-20
Edit paragraph in 2-contributing.asc (06-github)

rmzelle:
I originally tripped over "the change we would like to request", which I think is inaccurate/vague. The reason of a pull request is to request the acceptance of a change (where we're the ones providing the change), not to request a change (where the owner of the original project would have to make this change). The rest of the paragraph was also a bit repetitive, so I tried to improve that as well.


ben:
:heart:

rmzelle:
:sparkles:

PR #379, opened 2015-08-24
Nearly every other [source,console] section was preceded by a colon

crd:
Trivial change: I noticed a few paragraphs describing [source,console] blocks that were not preceded by colon.


ben:
Nice, thanks!

PR #381, opened 2015-08-25
"git log -1 HEAD" gets you the latest commit in the current branch.  …

srwhite:
…This isn't helpful if someone pushes to another branch.

### DIFF
--- a/book/08-customizing-git/sections/hooks.asc
+++ b/book/08-customizing-git/sections/hooks.asc
@@ -50,7 +50,7 @@ If this script exits non-zero, Git aborts the commit process, so you can use it 
 In the last section of this chapter, We'll demonstrate using this hook to check that your commit message is conformant to a required pattern.
 
 After the entire commit process is completed, the `post-commit` hook runs.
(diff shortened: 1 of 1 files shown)


ben:
Post-commit runs [at the end of `git commit`](http://git-scm.com/docs/githooks#_post_commit), which will always create a commit on top of HEAD. It won't fire when you're fetching or pulling. So `log -1 HEAD` is exactly what you want in a post-commit hook.

srwhite:
On 25/08/15 20:38, Ben Straub wrote:
> Post-commit runs at the end of |git commit|
> <http://git-scm.com/docs/githooks#_post_commit>, which will always
> create a commit on top of HEAD. It won't fire when you're fetching or
> pulling. So |log -1 HEAD| is exactly what you want in a post-commit hook.

Thanks, yes sorry, one of our sysadmins had copied this and used in
post-receive which is obviously different.  And post-receive gets useful
information on stdin as well.

-- 
Stephen White, CoreFiling Limited
http://www.corefiling.com
Phone: +44-1865-203192 ext. 564

Issue #365, opened 2015-08-19
core.editor documentation windows.


un1versal:
This is for discussion only to determine whats the best way to proceed.

Following up on https://github.com/git-for-windows/git/issues/278 where currently there is a issue of edits being terminated prematurely before their concluded while on Windows with notepad++ as a default editor not set as core.editor with ```-multiInst```

I had a look at [core.editor](https://github.com/progit/progit2/blob/master/book/08-customizing-git/sections/config.asc#coreeditor) (presuming this is right place to document this? Any suggestions?

The suggestion is to add some note to documentation on how to workaround the issue which seems to be resolved  by adding/settings Notepad++ with ```-multiInst``` until this is resolved in core and the installer detects and sets the editor properly as to avoid the issue in first place.

The reason this would be best in meanwhile is because https://github.com/msysgit/msysgit/issues/218 is now over a year old and no solution exists in core atm.

x86
```
 git config --global core.editor "'C:/Program Files/Notepad++/notepad++.exe' -multiInst" 
```
x64
```
 git config --global core.editor "'C:/Program Files (x86)/Notepad++/notepad++.exe' -multiInst" 
```

I dont mid making the edit and submitting the PR if someone points me to right page in case or suggest a better alternative or even additional documentation.

YueLinHo:
Perhaps improve [this section: First-Time Git Setup, Your Editor](https://github.com/progit/progit2/blob/master/book/01-introduction/sections/first-time-setup.asc#your-editor)
or [this section: Git Configuration, Basic Client Configuration, core.editor](https://github.com/progit/progit2/blob/master/book/08-customizing-git/sections/config.asc#coreeditor)

uNiversaI:
Does anyone else have any more feedback? Should I go ahead and make a PR adding this info to documentation or not?

YueLinHo:
> Should I go ahead and make a PR adding this info to documentation or not?

Submit a PR. Go! Go! Go! :smiley:

uNiversaI:
There we go https://github.com/progit/progit2/pull/382

un1versal:
I guess this is pretty much dead here, closing.

PR #382, opened 2015-08-25
[core.editor] improve first time setup to include Notepad++

un1versal:
On a Windows system, even though you have setup Notepad++ as default
system editor, Git seems to behave oddly during operations and you may trip over
glitches, like prematurely terminated Git operations before your edits
are complete.
Setting up core.editor in Git explicitly to Notepad++ fixes this behavior.

Related:

- [Preview of changes](https://github.com/uNiversaI/progit2/blob/a9f0bb3dca1cc28c13058012f2a6a47c10adbb47/book/01-introduction/sections/first-time-setup.asc)
- [Discussion issue #365](https://github.com/progit/progit2/issues/365)

### DIFF
--- a/book/01-introduction/sections/first-time-setup.asc
+++ b/book/01-introduction/sections/first-time-setup.asc
@@ -39,7 +39,8 @@ Many of the GUI tools will help you do this when you first run them.
 ==== Your Editor
 
 Now that your identity is set up, you can configure the default text editor that will be used when Git needs you to type in a message.
-If not configured, Git uses your system's default editor, which is generally Vim.
+If not configured, Git uses your system's default editor, which is system dependant.
+
 If you want to use a different text editor, such as Emacs, you can do the following:
 
 [source,console]
@@ -47,11 +48,29 @@ If you want to use a different text editor, such as Emacs, you can do the follow
 $ git config --global core.editor emacs
 ----
 
+While on a Windows system, if you want to use a different text editor, such as Notepad++, you can do the following:
+
+On a x86 system
+[source,console]
+----
+$ git config --global core.editor "'C:/Program Files/Notepad++/notepad++.exe' -multiInst -nosession" 
+----
+On a x64 system
+[source,console]
+----
+$ git config --global core.editor "'C:/Program Files (x86)/Notepad++/notepad++.exe' -multiInst -nosession" 
+----
+
+[NOTE]
+====
+Vim, Emacs and Notepad++ are popular text editors often used by developers on Unix based systems like Linux and OS X or a Windows system.
+If you are not familiar with either of these editors, you may need to search for specific instructions for how to set up your favorite editor with Git.
+====
+
 [WARNING]
 ====
-Vim and Emacs are popular text editors often used by developers on Unix based systems like Linux and Mac.
-If you are not familiar with either of these editors or are on a Windows system, you may need to search for instructions for how to set up your favorite editor with Git.
-If you don't set an editor like this and you don't know what Vim or Emacs are, you will likely get into a really confusing state when they are launched.
+You may find, if you don't setup an editor like this, you will likely get into a really confusing state when they are launched.
+Such example on a Windows system may include a prematurely terminated Git operation during a Git initiated edit.
 ====
 
 ==== Checking Your Settings


ben:
So Notepad++ is a fantastic editor, but it doesn't come in the box with Git or Windows, so I hesitate to recommend it like this. We don't want the user to have to install 20 things just to follow along with the book. However, given how alien things like Vim and Emacs are on Windows, we need to do something.

What about [GitPad](https://github.com/github/GitPad)? Does that play well with Notepad++? If so, it essentially solves the entire problem: install GitPad, set your favorite editor to be the default for `*.txt` files, and you're done. Plus we don't have to play favorites with Windows editors.

uNiversaI:
> We don't want the user to have to install 20 things just to follow along with the book. However, given how alien things like Vim and Emacs are on Windows, we need to do something.

OK.... but....



> What about GitPad? Does that play well with Notepad++? If so, it essentially solves the entire problem: install GitPad, set your favorite editor to be the default for *.txt files, and you're done. Plus we don't have to play favorites with Windows editors.

Well If I install GitPad, that is one more program I have to install that I dindt have before or never used before, so it interesting you first say you dont want people to install 20 things but then find it acceptable if they install 19.

Sorry your argument makes little sense to me.

ben:
So yeah, I changed my mind in the middle of that comment. It kind of hinges on how many recommendations we're making, and if the only example of configuration on Windows is for an editor the user doesn't have, that's confusing.

Part of the problem here is the way system-wide editors are configured. If you want to change it in a Unix, you set `$EDITOR`, which Git will honor. If you want to do the same in Windows, you change the editor associated with `.txt` files, but Git will still use vim. Installing GitPad bridges this gap, and I tend to think that's the way it should work (though I realize others disagree with me).

Here are some options I see:

* Outline the configuration for Vim or Emacs in the main text, and in a `[NOTE]` section mention Windows configuration, and cover both Notepad++ and GitPad. 
* Make a note on GUI editors in general, note that there might be some extra command-line options required to make Git wait for the editor to finish. Mention that the user may want to look into GitPad if they're on Windows.

Does one of those sound okay? Can you think of something better?

uNiversaI:
Well, Ill think on this a bit, I hesitate on having to install gitpad + notepad++ or whatever, learn how to that works and document it, but OK, Ill do that also in addition.

However, you cant just document programs which are likely to be used in one particular OS, its a pain to have fragmentation in the tools and methods used, but when the core cant handle this in any ideal fashion then you must resort to whatever the path of least resistance is. For most users its just installing the obvious tools...

For that reason I will try to include your suggestions just give me some time.

uNiversaI:
Apparently Github Desktop uses GitPad but why if it doesn't work really. (I asked Gihub Support), news to me having dug up the location of this executable now hunting to find the configuration file, This is turning into a scavenger hunt and here was I thinking this was a simple job...

uNiversaI:
@ben see https://github.com/github/GitPad/issues/24

Well Im definitely not adding GitPad instructions to this, so it is what it is here until that is resolved.

Its very frustrating, on one hand I have your commnets and on the other Github Desktop App which uses GitPad still needs these core-editor additions else GitPad via GitHub Desktop does sod all and I dont have the skills required or years left in me to learn them.

ben:
Yeah, I see your point. I didn't realize GitPad didn't solve the Notepad++ problem.

We definitely want to document the Vim/Emacs path, since those tools come with Git even on Windows, and give the user some guidance on how to use their favored GUI editor as well. Since this is especially tricky on Windows, we should call it out with a `[NOTE]` section that mentions some of the difficulties, and shows the correct configuration for Notepad++, Sublime Text, and maybe one of the [other popular choices](https://help.github.com/articles/associating-text-editors-with-git/)? What do you think of that approach?

uNiversaI:
vim or emacs is already documented just needs adjusting the OS warning. Ill adjust this PR to make a note and call attention to the windows issue, and later when I have more time will add sublime text to it as I would have to go find out what it supports in order to document it.

Boy oh boy, do I regret using Windows for 5 minutes. or what.

Once Im done Ill ping you.

nebasuke:
Also, for people who are using Windows AND Cygwin, which I assume is relatively common when they're using git from the command-line, it might be best to make a note that the stated solution does not work out of the box from Cygwin.

Instead to fix the path problems in Cygwin you will need something like the following: 
'C:/Program Files (x86)/Notepad++/notepad++.exe' -multiInst -nosession "$(cygpath -w "$*")"

uNiversaI:
If anyone knows what to add for sublime text let me know, I dont have time to go learn about some tool I never used to document something here.

Else Ill just document the parts which resolve the issue for users such as myself at this time.

un1versal:
@ben 

I hope this is acceptable.

I also hope Im not expected to document every possible editor and cover scenarios that didn't apply this issue specifically, as that is well outside the scope of this PR.
That was never my intention with this PR, just to make it clear, I wanted to document and improve the immediately available docs to include a solution for the issue.

ben:
Yeah, let's go ahead with this. It's definitely better, and it doesn't make it hard to include instructions for even more editors. Thanks for this!

PR #385, opened 2015-08-27
Replace "Chapter 2" to reference to viewing-history.asc

satob:
Replace fixed string "Chapter 2" to reference to viewing-history.asc like <<_viewing_history>>.


satob:
Thank you for your suggestion. I've updated the reference.

jnavila:
Good catch! Thanks! :tada:

PR #386, opened 2015-08-27
plain-text standardized

devwebcl:
similar to the word "e-mail", in the book appears the word "plain-text" in different ways:

plaintext
plain text
plain-text

I just changed the only plaintext.


ben:
👍🏼

PR #392, opened 2015-08-29
Added missing shell prompts to console blocks in git-daemon.asc

crd:
This was the only issue I found when reviewing console blocks in Chapter 4. 

Hey @ben is this the pattern you prefer or would you rather fewer, larger PRs?


ben:
This actually works great. It's easy to review small diffs. Thanks!

PR #396, opened 2015-08-30
revision-selection.asc: Small typographic change

xifedI:
The fourth dot isn't part of the SHA-1, but ends the sentence.


ben:
Nice catch. Thanks!

PR #399, opened 2015-09-01
Fix console blocks in 08-customizing.git

crd:
This PR takes care of Chapter 8.

### DIFF
--- a/book/08-customizing-git/sections/attributes.asc
+++ b/book/08-customizing-git/sections/attributes.asc
@@ -121,7 +121,7 @@ $ git config diff.exif.textconv exiftool
 
 If you replace an image in your project and run `git diff`, you see something like this:
 
-[source,console]
+[source,diff]
 ----
 $ diff --git a/image.png b/image.png
 index 88839c4..4afcb7c 100644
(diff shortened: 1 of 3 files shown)


book/08-customizing-git/sections/attributes.asc
```diff
@@ -121,9 +121,9 @@ $ git config diff.exif.textconv exiftool
 
 If you replace an image in your project and run `git diff`, you see something like this:
 
-[source]
+[source,console]
 ----
-diff --git a/image.png b/image.png
+$ diff --git a/image.png b/image.png
```
ben: This isn't actually a command line, that's the output diff spits out. You might want to mark this block as `[source,diff]` if that works.

crd: I was surprised to see Pygments does indeed support diff -- I'll give that a shot, thanks.

crd:
Yeah, I noticed that right after I fired off the commit. https://github.com/crd/progit2/commit/55c59158997f52cd823694509fe1d10e23cc7440 should have taken care of that. Thanks @ben.

PR #405, opened 2015-09-03
remove unneeded double whitespaces

Lazar95:
Found them with a regexp '[a-z]  [a-z]' and hand-picked them so
the command line outputs don't get affected.


ben:
Nice. Thanks!

PR #408, opened 2015-09-04
Bold and italic follow Asciidoc semantics

Lazar95:
The book used a combination of Markdown and Asciidoc syntax for
bold and italic text. In the documentation, it says that * are
used for highlighting the whole word or phrase, and ** are used
to highlight a part of the word. The render still worked fine
before, but the source-code now follows the semantics.

http://asciidoctor.org/docs/user-manual/#bold-and-italic

### DIFF
--- a/book/02-git-basics/sections/undoing.asc
+++ b/book/02-git-basics/sections/undoing.asc
@@ -128,6 +128,6 @@ Don't ever use this command unless you absolutely know that you don't want the f
 
 If you would like to keep the changes you've made to that file but still need to get it out of the way for now, we'll go over stashing and branching in <<_git_branching>>; these are generally better ways to go.
 
-Remember, anything that is __committed__ in Git can almost always be recovered.
+Remember, anything that is _committed_ in Git can almost always be recovered.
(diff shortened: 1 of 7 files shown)


jnavila:
We only highlight words. So this PR doesn't really make sense.

Lazar95:
The book uses `**` for highlighting whole words:

>     The first option ignores whitespace **completely** when comparing lines [...]

The Asciidoc documentation (my emphasis):

> To render a **word or phrase** with bold styling, place an asterisk (`*`) at the beginning and end of the text you wish to format. To bold **a letter or letters** in a string of text, place two asterisks (`**`) before and after the letter or letters.

As their example clearly states, `**` is used only for highlighting letters as parts of a word:

>     *charm*.
>     **u**ltimate

As you said, this book highlights only words and phrases, so it should use `*` instead of `**`.

I assume that the `**` syntax this book uses is a leftover from the first edition when the book was written using Markdown because using `*` in Markdown renders italic and `**` renders bold. In Asciidoc, `*` is used for bold words and phrases and `_` is used for italic words and phrases.

PR #412, opened 2015-09-11
import-hg.asc: change order of lines

xifedI:
The modified `/tmp/authors` file has the last two lines swapped compared to the original file.


ben:
:+1:

PR #413, opened 2015-09-12
transfer-protocols.asc: Update the git-receive-pack response explanation

xifedI:
Update the explanation to match the response shown in the console block.


ben:
Nice catch!

PR #414, opened 2015-09-15
Fix example output of "git clone"

moy:
"git clone" does not say "Initialized empty Git repository" (it may have
done so in very old versions), but "cloning into '$repo'". Reflect this
in the examples.

Noticed by Tim Castelijns:
http://stackoverflow.com/questions/32580578/why-the-git-commit-am-command-is-working-with-untraked-file-in-the-documenta


ben:
Nice catch. Thanks!

PR #416, opened 2015-09-15
Update hooks.asc

ajoslin103:
took me a few trial and errors to figure out what 'named appropriately' meant -- this would clear it up

also, you had the proper name of Git rather than the .git subdirectory


ben:
Nice, that's much clearer.

PR #424, opened 2015-09-27
Run pngquant + optipng

vmarkovtsev:
Optimize the size of PNG images by running the following two commands:

```
find . -name *.png -exec pngquant -f --skip-if-larger {} -o {} \;
find . -name *.png -exec optipng -o5 {} \;
```

These two commands save about 8.9M (64%). Evaluate the size with

```
find . -name *.png -print0 | xargs -0 du -csh
```

References:

* pngquant: https://pngquant.org
* optipng: http://optipng.sourceforge.net


ben:
Nice. Thanks!

jnavila:
If I recall, the PNGs from the repo aren't deployed on the web site. So, this only helps for ebooks and pdf.

PR #427, opened 2015-10-01
Updated installing.asc for Centos/Fedora

### LINKED ISSUE #425: For Centos, the installation section is missing 'perl-devel' package and workaround for 'docbook2X' package
I discovered some issues with the [installation instruction](https://github.com/progit/progit2/blob/master/book/01-introduction/sections/installing.asc) when compiling git on Centos 6.7. Namely `perl-devel` and `docbook2X` is missing from the `yum install` instruction. Additionally, some workarounds is needed for `docbook2X`.

# perl-devel

Without the package, I get this error when running `make`

    Can't locate ExtUtils/MakeMaker.pm

# docbook2X

Note that the X at the end is uppercase for Fedora/Centos, and lowercase for Ubuntu/Debian. Actually looking into it a bit further, it seems like it depends on which Centos repository you're using - it's uppercase in EPEL and lowercase in RepoForge. Also, EPEL and RepoForge are third-party repositories, I'm not sure if this package is in the default repositories.

Also, the package for Fedora/Centos has a different name for the binary which is `db2x_docbook2texi`. You'll need to create a symlink

    sudo ln -s /usr/bin/db2x_docbook2texi /usr/bin/docbook2x-texi

See [this](https://groups.google.com/forum/#!topic/git-users/DMaDpy8Bpww)

Again, this is using `docbook2X` from EPEL. I'm not sure if this issue applies if you use `docbook2x` from RepoForge.

i-give-up:
Compilation on Centos 6 requires perl-devel and docbook2X (uppercase X). I'm assuming this is the same for all other RHEL-likes.

docbook2X is only available through third-party repo like EPEL. Additionally, the binary is named `db2x_docbook2texi` instead of `docbook2x-texi` so symlinking is required.

Fixes #425.

### DIFF
--- a/book/01-introduction/sections/installing.asc
+++ b/book/01-introduction/sections/installing.asc
@@ -73,18 +73,27 @@ For example, if you're on a system that has yum (such as Fedora) or apt-get (suc
 [source,console]
 ----
 $ sudo yum install curl-devel expat-devel gettext-devel \
-  openssl-devel zlib-devel
+  openssl-devel perl-devel zlib-devel
 $ sudo apt-get install libcurl4-gnutls-dev libexpat1-dev gettext \
   libz-dev libssl-dev
 ----
 
-In order to be able to add the documentation in various formats (doc, html, info), these additional dependencies are required:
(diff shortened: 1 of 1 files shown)


ben:
Nice. Thanks!

PR #428, opened 2015-10-04
Give more details on Windows system-level config file.

tacker66:
Signed-off-by: Thomas Ackermann <th.acker@arcor.de>

### DIFF
--- a/book/01-introduction/sections/first-time-setup.asc
+++ b/book/01-introduction/sections/first-time-setup.asc
@@ -18,7 +18,9 @@ Each level overrides values in the previous level, so values in `.git/config` tr
 
 On Windows systems, Git looks for the `.gitconfig` file in the `$HOME` directory (`C:\Users\$USER` for most people).
 It also still looks for `/etc/gitconfig`, although it's relative to the MSys root, which is wherever you decide to install Git on your Windows system when you run the installer.
-However, if you are using Git for Windows 2.x, it is `C:\Documents and Settings\All Users\Application Data\Git\config` on Windows XP, and it is `C:\ProgramData\Git\config` on Windows Vista and newer.
+If you are using Git for Windows 2.x or later, there is also a system-level config file at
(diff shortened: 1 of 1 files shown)


book/01-introduction/sections/first-time-setup.asc
```diff
@@ -18,7 +18,9 @@ Each level overrides values in the previous level, so values in `.git/config` tr
 
 On Windows systems, Git looks for the `.gitconfig` file in the `$HOME` directory (`C:\Users\$USER` for most people).
 It also still looks for `/etc/gitconfig`, although it's relative to the MSys root, which is wherever you decide to install Git on your Windows system when you run the installer.
-However, if you are using Git for Windows 2.x, it is `C:\Documents and Settings\All Users\Application Data\Git\config` on Windows XP, and it is `C:\ProgramData\Git\config` on Windows Vista and newer.
+However, if you are using Git for Windows 2.x, there is additionally a config file on Windows system level in
+`C:\Documents and Settings\All Users\Application Data\Git\config` on Windows XP, and in `C:\ProgramData\Git\config` on Windows Vista and newer.
+This config file can only be changed by `git config -f <file>`.
```
ben: Let's reword this a bit. 
> If you are using Git for Windows 2.x or later, there is also a system-level config file at `C:\Documents and Settings\All Users\Application Data\Git\config` on Windows XP, or `C:\ProgramData\Git\config` on Windows Vista and newer. This config file can only be changed by `git config -f <file>`.

By the way, is that last line correct? You can't launch wordpad as an admin to edit it?

tacker66: Thanks for the rewording!

You are of course right about using notepad as an admin to edit this config file,
but as we didn't tell the user at all about the possibility to directly edit config files
with notepad in this section only the fact that you have to be admin to call
'git config' successfully  in this case should be added.

ben: Yeah, I guess that fits in then. Power users will figure out how to `sudo vi` the file. Thanks!

PR #430, opened 2015-10-08
Fix typos in Chapter 6

kdelwat:
Fixes two minor typos in chapter 6, the first in the Maintaining section and the second in the Scripting section.


ben:
Nice. Thanks!

PR #432, opened 2015-10-14
Improve Smart HTTP section

### LINKED ISSUE #122: book errata
In section 4.6 (Smart HTTP Server Setup) You have an error:

This code specifies instructions for the auth_basic modual for apache (which provides AuthType Basic):

&lt;LocationMatch "^/git/.*/git-receive-pack$"&gt;
    AuthType Basic
    AuthName "Git Access"
    AuthUserFile /opt/git/.htpasswd
    Require valid-user
"&lt;/LocationMatch"&gt;

But this instruction just below it is for use with the auth_digest module (Providing AuthType Digest);

htdigest -c /opt/git/.htpasswd "Git Access" schacon

The second element of code should be
htpasswd ...   instead of htdigest.

vitalyq:
Issues arise during the very basic setup of Smart HTTP using the tutorial.

1. User files generated with htdigest and htpasswd are not compatible.
Thus, if AuthType Basic is used, htdigest must be changed to htpasswd.
The fix also resolves #122.

2. Apache process doesn’t have write permissions for the /opt/git/ subdirectories.
To address this, I suggest restoring the paragraph from this section:
https://git-scm.com/book/en/v1/Git-on-the-Server-Public-Access.
Namely, `$ chgrp -R www-data /opt/git`.

3. Authentication fails during push.
The initial ref advertisement request will fail without even giving an opportunity for authentication.
This is described in the git-http-backend man page examples: https://www.kernel.org/pub/software/scm/git/docs/git-http-backend.html
I have updated and simplified the authentication paragraph from the example to fit the context.
Also, the paragraph is merged with the git-core access authorization paragraph for brevity.

### DIFF
--- a/book/04-git-server/sections/smart-http.asc
+++ b/book/04-git-server/sections/smart-http.asc
@@ -18,6 +18,13 @@ $ a2enmod cgi alias env
 
 This also enables the `mod_cgi`, `mod_alias`, and `mod_env` modules, which are all needed for this to work properly.
 
+You’ll also need to set the Unix user group of the `/opt/git` directories to `www-data` so your web server can read- and write-access the repositories, because the Apache instance running the CGI script will (by default) be running as that user:
+
+[source,console]
+----
+$ chgrp -R www-data /opt/git
+----
+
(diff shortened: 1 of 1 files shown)


ben:
I like it, thanks!

PR #434, opened 2015-10-15
Fix filename inconsistency believed to cause broken website image

codemagician:
I noticed a broken link on the live website.  The image uses an underscore character, whereas the other files use the hyphen character as a name separator.

screen shot 2015-10-15 at 13 23 58

Note, this does not fix the broken image on the website, but I am making the assumption that the reason the image is missing is that someone or some automatic deploy tool missed the file because it contained an underscore and not a hyphen?  This is just a guess.


ben:
*Sigh.* This has been an ongoing problem, which @schacon and others who have access to the deployment of the git-scm.com site haven't been able to address. See https://github.com/git/git-scm.com/pull/505.

jnavila:
As ben said. It's already fixed in the book (check with the pdf) but could not reach the website. Closing as it is related to git-scm.

PR #435, opened 2015-10-18
change 'winstore' credential option to 'wincred'

mmikeww:
I can't seem to find any information on using 'winstore' as a credential saving option on Windows. The link is outdated. 

With the default Git for Windows 2.6.1, when I run `git help -a` like the docs tell you to, it shows "crendential-wincred". 
https://git-scm.com/docs/gitcredentials

GitHub recommends to use "wincred" as well:
https://help.github.com/articles/caching-your-github-password-in-git/


ben:
Nice, wincred looks like an improvement. Thanks!

PR #436, opened 2015-10-18
[RFC] Add companions packages of git for linux installs

jnavila:
On most linux distribution, the git package only provides the bare
command line git command. Just adding gitk and git-gui is the least.

I think that if we really wanted to provide the same level of features as the self packaged Git versions, we should even make people install the meta-package "git-all", that pulls git-svn, git-cvs, git-p4, gitweb and git-email.


ben:
This is a good change, I like it. Should we include a note about the `git-all` package then?

jnavila:
As an afterthought, I think the simplest is just to make them install git-all.

Issue #422, opened 2015-09-24
zsh chapter enhancements


danielshahaf:
Two enhancements to the zsh chapter:

1. zsh ships with a git completion that has features the one you describe lacks.  For example, zsh's git completion completes the `cat-file` command, doesn't offer unmodified files for `git add`, and so on. It requires no installation. It should at least be mentioned as an alternative to the zsh git completion that ships with git.

2. zsh ships with a framework for getting information from version control system, called [vcs_info](http://zsh.sourceforge.net/Doc/Release/User-Contributions.html#Version-Control-Information) (by @ft), which can be used by adding a couple of lines to .zshrc without requiring a full-fledged third-party framework. A minimal example is: `autoload -Uz vcs_info; precmd() { vcs_info; [[ -n ${vcs_info_msg_0_} ]]  && print -Plr - $vcs_info_msg_0_ $vcs_info_msg_1_ }`

ben:
Wow, that sounds like great content. Want to write up a pull request? It'd be nice if you got credit for this.

danielshahaf:
Sure, I could give it a shot.  However, the wording in CONTRIBUTING.md is unclear about how contributions should be licensed.  Would it be OK if my PR were MIT-licensed?

ben:
Totally not a lawyer here. The work itself is CC-BY-NC-SA, which requires attribution and non-commercial uses, and we also require license to make a third-edition print version.

I *think* licensing your commits under MIT is compatible with those goals, but I'm nowhere *near* sure about that. How would you mark them as MIT? Add the license in the commit message?

danielshahaf:
> How would you mark them as MIT? Add the license in the commit message?

Sure, that's a good idea.

danielshahaf:
Ping? What license should I put my patch under for it to be acceptable?  I don't really mind which license, so long as I know in advance which one it is.

The only reason I'm asking for you to specify a license in advance is that your CONTRIBUTING.md doesn't do that.  CONTRIBUTING.md asks contributors to license their patches under licenses to be determined _after the fact_, _by a third party_.  No other project (open source or otherwise) does that.

(Sorry for the headache — I don't like having to care about such details any more than you do..)

ben:
That's true. It seems others either haven't noticed it, or don't mind. I'll look into what to do about this long term, but in the meantime MIT should be fine.

danielshahaf:
Thanks!  Pull request sent.

PR #440, opened 2015-10-27
zsh += vcs_info and native completion

danielshahaf:
Patch for [#422](https://github.com/progit/progit2/issues/422#issue-108144184)

(As discussed in #422, this patch is made available under the MIT license.)


ben:
Nice. Thanks!

Issue #438, opened 2015-10-27
Mistake in figure 3-9 - swapped branches


ZvikaZ:
There is a mistake in figure 3-9.
According to previous figures, and to the code just below it, 87ab2 should be 'testing' and  c2b9e should be 'master'

crd:
This issue is a duplicate of https://github.com/progit/progit2/issues/136 which was subsequently fixed with https://github.com/progit/progit2/commit/87254184fd3642148c284df68c59ae20751ca44a. It seems like   the outdated html content at git-scm.com is the source of the majority of issues opened against progit2.

YueLinHo:
So, this issue should be closed as duplicate.

PR #441, opened 2015-10-31
Fixed image 3-9 for issue #438. I hope this helps others.

StevePrice:
This should help.


ben:
[Created](http://forum.atlas.oreilly.com/t/angle-brackets-inside-source-block-are-replaced-with-entities-html-only/254). Since this commit isn't technically wrong in the source, and the other commit is a no-op, I'm going to close this out. Thanks for helping out!

PR #450, opened 2015-11-17
Use quotes - Single quotes not platform independent

AndyGee:
Using single quotes results in the error - fatal: Failed to resolve '[tag name]'' as a valid ref.


jnavila:
Works for me. Which version of git on which platform are you using?

AndyGee:
Git v2.6.3.windows.1 on Windows 10 - Using double quotes also works on Ubuntu + Mac

ben:
:+1:

PR #452, opened 2015-11-18
git diff: --cached → --staged

hginzel:
To be consistent with term “staging area”.

### DIFF
--- a/book/02-git-basics/sections/recording-changes.asc
+++ b/book/02-git-basics/sections/recording-changes.asc
@@ -354,11 +354,11 @@ index 643e24f..87f08c8 100644
 +# test line
 ----
 
-and `git diff --cached` to see what you've staged so far (--staged and --cached are synonyms):
+and `git diff --staged` to see what you've staged so far (--cached and --staged are synonyms):
 
 [source,console]
 ----
-$ git diff --cached
+$ git diff --staged
 diff --git a/CONTRIBUTING.md b/CONTRIBUTING.md
 index 8ebb991..643e24f 100644
 --- a/CONTRIBUTING.md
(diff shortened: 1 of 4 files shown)


ben:
I like the change from `--cached` to `--staged`, but it's the prose changes that are causing more thought. Is this everywhere we mention the index? I think it's valuable to mention that "staging area" and "index" mean the same thing, since a user will run into both of those in the man pages. But being consistent is good, too.

book/02-git-basics/sections/recording-changes.asc
```diff
@@ -506,7 +506,7 @@ Changes to be committed:
 ----
 
 The next time you commit, the file will be gone and no longer tracked.
-If you modified the file and added it to the index already, you must force the removal with the `-f` option.
+If you modified the file and added it to the staging area already, you must force the removal with the `-f` option.
```
YueLinHo: perhaps using:
```
[...] added it to the index(staging area) already, [...]
```

ben:
@ginzel, are you still interested in getting this merged?

aollier:
The `SYNOPSIS` part of the [documentation](https://git-scm.com/docs/git-diff "git diff documentation") showcases the `--cached` option. We can read in the `DESCRIPTION` part that:
> --staged is a synonym of --cached.

So from this point of view, I think we should let the `--cached` option. Then the reader won't be confused when reading the documentation.

aizulaiz:
update

aollier:
@ben I think this pull request is not a good idea as I explained in my previous comment and @hginzel does not answer.
Do you agree to close it ?

ben:
Agreed. Closing this one out.

PR #460, opened 2015-11-23
Fix a grammatical mistake in submodules.asc

brettcannon:


ben:
Well, you're right that there are issues with how this reads, but I'm not sure your fix is the best one. Really what we want to say is something like "if you make changes in this state, even if you make commits, they're not _really_ being tracked, and will most likely be lost when you update your submodule again." But that's not the right way to say it. Any ideas?

brettcannon:
I've taken another stab and trying to explain the situation.

ben:
:sparkles:

PR #461, opened 2015-11-24
Correct P4Merge link in "External Merge and Diff Tools"

al-the-x:
The current link is 404. I also reworded the text to convey the intent of the link, but I'm happy to adjust if there are objections.


ben:
The original text was intended for print, so we just spelled out the URL, but in all the formats this will ship to, your patch is better. *Especially* since the URL is uglier now. Thanks!

al-the-x:
> _Especially_ since the URL is uglier now. 

_Now with 100% more Google Juice!_ The term _permalink_ is lost on marketing departments... Glad to contribute!

PR #464, opened 2015-12-01
"pointer at" to "pointer to"

qpid:
The rest of Section 3.1 describes pointers as pointing 'to' something, and on this particular line it was 'creates a new pointer at', which might make the sentence seem a little incoherent on first glance.


ben:
Nice. Thanks!

PR #471, opened 2015-12-17
fix a index for "C" in libgit2.asc

satob:
There is unnecessary copyright mark at the beggining of "A2.2 Embedding Git in your Applications - Libgit2".
https://git-scm.com/book/en/v2/Embedding-Git-in-your-Applications-Libgit2
This is because of (((C))) in libgit2.asc is treated as a copyright mark.


ben:
Nice catch. Thanks!

PR #477, opened 2015-12-24
Update transfer-protocols.asc

yann-soubeyrand:
I think there is a little mistake ;-)

### DIFF
--- a/book/10-git-internals/sections/transfer-protocols.asc
+++ b/book/10-git-internals/sections/transfer-protocols.asc
@@ -199,7 +199,7 @@ The connection is initiated with this request:
 ----
 
 That's the end of the first client-server exchange.
-The client then makes another request, this time a `POST`, with the data that `git-upload-pack` provides.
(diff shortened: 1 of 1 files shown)


ben:
```shell
$ git-send-pack
bash: git-send-pack: command not found
$ git-upload-pack
usage: git upload-pack [--strict] [--timeout=<n>] <dir>
```

Looks like the typo is the other way around. Want to change the spot where `git-send-pack` is mentioned to read `git-upload-pack` instead?

yann-soubeyrand:
I think this is really git-send-pack:
~~~
$ apropos git-send-pack
git-send-pack (1)    - Push objects over Git protocol to another repository
~~~
The point is that the git-upload-pack command exists on its own (like git-receive-pack) but send-pack (like fetch-pack) are git subcommands. I did a new commit to take this into account and it seems to be more coherent with the rest of the section.

ben:
Alright. I'm cool with it. Thanks!

PR #479, opened 2015-12-30
Update Ch8 to refer to .gitattributes consistently

crd:
All modifications to the .gitattributes file are now illustrated using `[source,ini]` blocks. Previously it was a mishmash of `[source,ini]` and `[source,console] $echo foo >> .gitattributes` commands.


ben:
:+1: Thanks!

PR #480, opened 2015-12-30
Replace hardcoded incorrect reference to Chapter 6

crd:
Updated policy.asc in Chapter 8 to use a link to the relevant section in Chapter 7 rather than a hardcoded and incorrect reference to Chapter 6.

### DIFF
--- a/book/08-customizing-git/sections/policy.asc
+++ b/book/08-customizing-git/sections/policy.asc
@@ -427,7 +427,7 @@ target_shas.each do |sha|
   end
 end
 ----
-This script uses a syntax that wasn't covered in the Revision Selection section of Chapter 6. You get a list of commits that have already been pushed up by running this:
+This script uses a syntax that wasn't covered in the <<_revision_selection>> section of <<_git_tools>>. You get a list of commits that have already been pushed up by running this:
 
 [source,ruby]
 ----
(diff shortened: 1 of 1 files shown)


book/08-customizing-git/sections/policy.asc
```diff
@@ -427,7 +427,7 @@ target_shas.each do |sha|
   end
 end
 ----
-This script uses a syntax that wasn't covered in the Revision Selection section of Chapter 6. You get a list of commits that have already been pushed up by running this:
+This script uses a syntax that wasn't covered in the <<_revision_selection>> section of <<_git_tools>>. You get a list of commits that have already been pushed up by running this:
```
ben: I think this will render weirdly. Let's just do this:

> This script uses a syntax that wasn't covered in <<_revision_selection>>.

I think that'll render better in all formats.

crd:
My build environment produces output files that look a bit different from the official builds:

image

In the official .pdf I see that <<_git_tools>> renders correctly as "Chapter 7" [see page 195 of the official .pdf](https://progit2.s3.amazonaws.com/en/2015-12-24-7bc46/progit-en.959.pdf):

image

That said, I'm happy to make the change you recommend.

crd:
Removed superfluous reference to Chapter 7 as per @ben's recommendation.

ben:
:tada:

PR #481, opened 2015-12-30
Fix 'dependant' typo in first-time-setup.asc

crd:
Deleted a redundant sentence fragment rather than directly fix the typo.


ben:
Yeah, this is better. Thanks for all of these!

PR #482, opened 2015-12-30
Capitalize Version Control System for consistency

crd:
To be consistent with Chapter 1, capitalize the rest of the occurrences of "centralized version control system" and "version control system" throughout the rest of the text.


ben:
Not so sure about this one. The terms aren't _proper_ names, so the only time I'd think it was proper would be just when the acronym is introduced, to reinforce the connection. I'm willing to be convinced, though.

PR #485, opened 2015-12-31
Use correct brackets to indicate mandatory/optional parameters

crd:
Square brackets [] indicate an optional parameters -- use the right ones!


ben:
:+1:

PR #488, opened 2016-01-03
Strip "origin/" prefix from branch name

mkarg:
git svn clones prefixes all SVN branch names by "origin/" in git, which is not wanted when migrating from svn to git. Hence, the prefix has to be stripped, simply in the same way it is done for tags already.


ben:
Ah, I see. Nice catch!

PR #489, opened 2016-01-03
Remove duplicate 'trunk' branch

mkarg:
svn git creates 'trunk' branch which points to the same commit as 'master' which is simpy misleading. Let's just remove it!

### DIFF
--- a/book/09-git-and-other-scms/sections/import-svn.asc
+++ b/book/09-git-and-other-scms/sections/import-svn.asc
@@ -92,7 +92,7 @@ $ rm -Rf .git/refs/remotes
 
 Now all the old branches are real Git branches and all the old tags are real Git tags.
 
-Unfortunately `git svn` creates one extra branch named `trunk` for svn's trunk, which is essentially the same as the already created `master` branch (it actually refers to the same commit). Hence, let's remove this doublet:
+There's one last thing to clean up. Unfortunately, git svn creates an extra branch named trunk, which maps to Subversion's default branch, but the trunk ref points to the same place as master. Since master is more idiomatically Git, here's how to remove the extra branch:
 
 [source,console]
 ----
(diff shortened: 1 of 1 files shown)


book/09-git-and-other-scms/sections/import-svn.asc
```diff
@@ -91,6 +91,14 @@ $ rm -Rf .git/refs/remotes
 ----
 
 Now all the old branches are real Git branches and all the old tags are real Git tags.
+
+Unfortunately `git svn` creates one extra branch named `trunk` for svn's trunk, which is essentially the same as the already created `master` branch (it actually refers to the same commit). Hence, let's remove this doublet:
```
ben: How about:

> There's one last thing to clean up. Unfortunately, `git svn` creates an extra branch named `trunk`, which maps to Subversion's default branch, but the `trunk` ref points to the same place as `master`. Since `master` is more idiomatically Git, here's how to remove the extra branch:

ben:
I like the change here, just trying to make the prose match its surroundings a bit better.

book/09-git-and-other-scms/sections/import-svn.asc
mkarg: Great! Go for it! :-)

book/09-git-and-other-scms/sections/import-svn.asc
```diff
@@ -91,6 +91,14 @@ $ rm -Rf .git/refs/remotes
 ----
 
 Now all the old branches are real Git branches and all the old tags are real Git tags.
+
+There's one last thing to clean up. Unfortunately, git svn creates an extra branch named trunk, which maps to Subversion's default branch, but the trunk ref points to the same place as master. Since master is more idiomatically Git, here's how to remove the extra branch:
```
ben: One last thing: can you set some of the "trunk"s and "master"s in `monospace`? Maybe like this:

```markdown
There's one last thing to clean up. Unfortunately, `git svn` creates an extra 
branch named `trunk`, which maps to Subversion's default branch, but the 
`trunk` ref points to the same place as `master`. Since `master` is more
idiomatically Git, here's how to remove the extra branch:
```

mkarg:
Good? :-)

ben:
:joy:

PR #491, opened 2016-01-03
Remove superfluous phrase.

### LINKED ISSUE #486: Odd (wrong?) topic sentence of subtree-merges
The topic sentence of the third paragraph of this section is "The idea of the subtree merge is that you have two projects, and one of the projects maps to a subdirectory of the other one and vice versa."

Are you sure that this is what you meant?  It seems to say "you have two projects and one of them maps to a subdirectory of the other, while the other maps to a subtree of the first."

The meaning of "vice versa" here is not at all clear.  Possibly it would be best to cut the "and vice versa," since I think usually you have a project that includes a second project as subdirectory (for example, because the second project provides a library).

rpgoldman:
The superfluous "and vice versa" is incorrect here -- the two projects
/don't/ each have the other as subdirectories.  In fact only one has the
other as a subdirectory.

Fixes #486.

### DIFF
--- a/book/07-git-tools/sections/subtree-merges.asc
+++ b/book/07-git-tools/sections/subtree-merges.asc
@@ -1,7 +1,7 @@
 [[_subtree_merge]]
 ===== Subtree Merging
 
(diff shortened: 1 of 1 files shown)


ben:
:heart:

PR #492, opened 2016-01-03
git push --all does not push tags

mkarg:
It is incorrect that git push origin -all would push tags on the git server. Actually you have to do git push --tags to also push the tags.


ben:
Nice catch, thanks!

PR #496, opened 2016-01-09
Make output of `git log` consistent with preceding figure

crd:
Make commit messages consistent with the preceding figure and correct the `(HEAD -> master, testing)` `--decorate` output.


ben:
Good eyes! :eyes:

PR #499, opened 2016-01-12
Make examples that refer to files by name consistent

crd:
Most of the time, file names are marked up like so: `filename.txt` -- in a few places there is no markup and in others it's done in this way: ``filename.txt''. This commit makes them all consistent.


ben:
:+1: Thanks!

PR #505, opened 2016-01-18
avoid fetching subtree's tags into project's tagspace

bronson:
Most people won't want to intermingle the subtree project's tags with their own.


ben:
:+1: Thanks!

Issue #511, opened 2016-01-26
warning for git commit -a -m""


rodsouza9:
The thing is almost all of you general readers are new to git. so The git commit -a -m when you teach how to skip the staging area you are putting the newbies at a risk for becoming lazy. As people new to git they should be diligent in their actions. I understand that everything is not needed to be staged, but at first it is good practice.

PR #512, opened 2016-01-26
Warning for git commit -a -m""

rodsouza9:
I know this is a very small change but i believe that it will be worth is also read the issue:
https://github.com/progit/progit2/issues/511

### DIFF
--- a/book/02-git-basics/sections/recording-changes.asc
+++ b/book/02-git-basics/sections/recording-changes.asc
@@ -466,7 +466,8 @@ $ git commit -a -m 'added new benchmarks'
 ----
 
 Notice how you don't have to run `git add` on the `CONTRIBUTING.md` file in this case before you commit.
-But beware because you can commit files that you don't necesarily want to commit.
+But beware because you can commit files that you don't necesarily want to commit. Be diligent when you need to commit files, and most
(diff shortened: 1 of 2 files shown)


rodsouza9:
also I apologies for the readme commits

book/02-git-basics/sections/recording-changes.asc
```diff
@@ -466,6 +466,8 @@ $ git commit -a -m 'added new benchmarks'
 ----
 
 Notice how you don't have to run `git add` on the `CONTRIBUTING.md` file in this case before you commit.
+But beware because you can commit files that you don't necessarily want to commit. Be diligent when you need to commit files, and most
+of the time you should stage and then commit you files.
```
ben: I like where this is going. How about something like this?

```
Notice how you don't have to run `git add` on the `CONTRIBUTING.md` file in this case before you commit.
That's because the `-a` flag includes all changed files.
This is convenient, but be careful; sometimes this flag will cause you to include unwanted changes.
```

rodsouza9: This is a much better way of saying it thank you.

ben:
Great, thanks!

PR #514, opened 2016-01-30
replace branch by checkout

migueldvb:
I think this is what is meant in the context of the HEAD reference.


ben:
Not quite, actually., `git branch newbranch` tries to create `refs/heads/newbranch` that points to the **current commit**, which is why it needs `HEAD`. Maybe the example isn't the most clear, but it's correct. Does that make sense?

PR #521, opened 2016-02-11
Small fixes

hedrok:
I've put several small fixes to one PR, I hope this will save your time.
If it's better to make separate PRs for them - please tell me.

### DIFF
--- a/book/10-git-internals/sections/objects.asc
+++ b/book/10-git-internals/sections/objects.asc
@@ -242,6 +242,8 @@ $ echo 'first commit' | git commit-tree d8329f
 fdf4fc3344e67ab068f836878b6c4951e3b15f3d
 ----
 
+You will get another hash value because of different creation time and author data.
+Replace commit and tag hashes with your own checksums further in this chapter.
 Now you can look at your new commit object with `cat-file`:
 
 [source,console]
--- a/book/10-git-internals/sections/objects.asc
+++ b/book/10-git-internals/sections/objects.asc
@@ -242,7 +242,7 @@ $ echo 'first commit' | git commit-tree d8329f
 fdf4fc3344e67ab068f836878b6c4951e3b15f3d
 ----
 
-You will get another hash value because of different creation time and author data.
+You will get a different hash value because of different creation time and author data.
 Replace commit and tag hashes with your own checksums further in this chapter.
 Now you can look at your new commit object with `cat-file`:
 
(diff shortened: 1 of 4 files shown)


book/10-git-internals/sections/objects.asc
```diff
@@ -242,6 +242,8 @@ $ echo 'first commit' | git commit-tree d8329f
 fdf4fc3344e67ab068f836878b6c4951e3b15f3d
 ----
 
+You will get another hash value because of different creation time and author data.
```
ben: Instead of "another", can you use "a different"? I initially read this as "you'll get a second hash value."

hedrok: Of course. Fixed.

ben:
Nice! Thanks!

book/10-git-internals/sections/packfiles.asc
```diff
@@ -118,8 +119,8 @@ Because you never added them to any commits, they're considered dangling and are
 The other files are your new packfile and an index.
 The packfile is a single file containing the contents of all the objects that were removed from your filesystem.
 The index is a file that contains offsets into that packfile so you can quickly seek to a specific object.
-What is cool is that although the objects on disk before you ran the `gc` were collectively about 22K in size, the new packfile is only 7K.
-You've cut your disk usage by ⅔ by packing your objects.
+What is cool is that although the objects on disk before you ran the `gc` were collectively about 15K in size, the new packfile is only 7K.
+You've cut your disk usage by ½ by packing your objects.
```
YueLinHo: Just a record: 2/3 -> 1/2 (Too small to read here)

hedrok: This is an intentional change. I've executed all commands in this chapter and measured sum size of all objects. It is 15K, not 22K. Probably author missed the fact that objects are compressed even when stored out of packfile...
I can provide script that executes all commands in this chapter till this line so that anyone can check my data.

@ben, maybe it would be better to just write "by half"?

YueLinHo: I checked the web page. It's good, not too small. :smile: 

image

ben: Yeah, "by half" is more idiomatic in English. You'd only use ½ for an exact fraction, and this isn't exact.

hedrok: Ok, I've created PR: #535

PR #522, opened 2016-02-12
Config commands were reversed

alexwidener:
x86 was where 64 should have been and vice versa.

### DIFF
--- a/book/01-introduction/sections/first-time-setup.asc
+++ b/book/01-introduction/sections/first-time-setup.asc
@@ -55,12 +55,12 @@ While on a Windows system, if you want to use a different text editor, such as N
 On a x86 system
 [source,console]
 ----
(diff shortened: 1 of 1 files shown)


ben:
:hushed: Nice catch!

ben:
Oh wait. No, it was already correct.

This has to do with the crazy way Windows decided to allow 32-bit programs to run on a 64-bit system. 32-bit programs are stored under `C:\Program Files` on a 32-bit system, and under `C:\Program Files (x86)` on a 64-bit system. That's because `C:\Program Files` on a 64-bit system is for 64-bit programs. Does that make sense?

Well, of course it doesn't. But do you understand it? :smile:

alexwidener:
Yeah, I got it. I was interpreting as if they were on the machine, not as if they were on two different OS systems. Silly. Thanks!

PR #529, opened 2016-02-17
Remove --detach from local-git-daemon.conf

sselberg:
--detach will fork and trigger an endless loop of respawns
If --detach is used you probably want to specify

expect fork[1]

[1] http://upstart.ubuntu.com/cookbook/#expect-fork


ben:
:+1:

Issue #426, opened 2015-09-30
Help with --ours merge strategy trick needed


hedrok:
Could someone please explain the last paragraph of book/07-git-tools/sections/advanced-merging.asc?..
I've tried a lot of variants, but however I merge the 'bugfix' branch to both 'master' and 'develop' branches, I get no conflict when I merge 'release' back to 'master'...
What should I do to get this conflict?..
---
This [--ours merge strategy] can often be useful to basically trick Git into thinking that a branch is already merged when doing a merge later on. For example, say you branched off a release branch and have done some work on it that you will want to merge back into your master branch at some point. In the meantime some bugfix on master needs to be backported into your release branch. You can merge the bugfix branch into the release branch and also merge -s ours the same branch into your master branch (even though the fix is already there) so when you later merge the release branch again, there are no conflicts from the bugfix.
---
I'm currently translating the book to Ukrainian and would like to fully understand everything I translate.

Thanks in advance!

mjg:
Before that paragraph, the book talks about resolving conflicts and how to set options for the default recursive merge strategy using '-X' (such as '-Xours') in order to get fewer conflicts.

In the paragraph you are referencing, the book talks about the "ours strategy" ('-Sours') which will never produce any conflicts because it simply takes "our side" as the merge result. In effect, this "marks the side branch as merged in" without even looking at its contents, and takes the contents of the main branch verbatim. No chance for conflicts there ;;)

hedrok:
Thank you very much for your time!

Yes, I understand that `-Xours` leads to no conflicts :)
I don't understand how the situation described in the quoted paragraph leads to conflict if we don't "merge -s ours the same branch into your master branch (even though the fix is already there)".
I'm trying to do the following (as I understand this paragraph):

```bash
# 1. prepare repository
git init .
echo '1' > a.txt
echo '2' >> a.txt
echo '3' >> a.txt
echo '4' >> a.txt
echo '5' >> a.txt
git add a.txt
git commit -m 'Initial commit'
# 2. > For example, say you branched off a release branch
git checkout -b release
# 3. > and have done some work on it that you will want to merge back into your master branch at some point.
sed -i 's/2/2 release/' a.txt
git commit -a -m 'release commit'
# 4. > In the meantime some bugfix on master needs to be backported into your release branch. making bugfix git checkout 
git checkout -b bugfix master
sed -i 's/5/5 bugfix/' a.txt
git commit -a -m 'bugfix commit'
git checkout master
git merge bugfix
[...]

PR #535, opened 2016-02-19
Use 'half' instead of '½'

hedrok:


ben:
:grinning:

hedrok:
Thanks for quick merge.

If you have some time, please take a look at #426
It still bothers me :)

PR #542, opened 2016-02-24
git-credential-read-only should match against username

jouve:
Hello,

In the current version, the first entry matching protocol/hostname and not protocol/username/hostname is returned.

For exemple:
.git-credentials:
```
http://titi:pwd1@exemple.com (1)
http://toto:pwd2@exemple.com (2)
```
then
```git clone http://toto@exemple.com``` will match (1) instead of (2).


ben:
Heyyy, nice catch. I like it, thanks!

Issue #541, opened 2016-02-24
Add TortoiseGit config for bugtraq and icon


YueLinHo:
I use TortoiseGit for my daily git work.

I want to add `.tgitconfig` and `Pro.ico` files into progit2 repository for these things:

**For [bugtraq] config:**

![default](https://cloud.githubusercontent.com/assets/5831804/13274637/ccc51626-dae7-11e5-8d63-ee8385e4e5d4.png)

**For icon config:**

![1](https://cloud.githubusercontent.com/assets/5831804/13274304/32984d86-dae5-11e5-98d1-405ee33b2020.png) (The source of that icon is made from `progit\epub\title.png`)

**May I add them?**

ben:
Yeah, that would be fine. I'm sure there are others that use these tools. Go ahead and submit a PR with before and after screen shots.

YueLinHo:
Fixed with PR #543.

PR #543, opened 2016-02-25
Fix issue #541: Add TortoiseGit config for bugtraq and icon

YueLinHo:
### Before

image

### After

image

### Before

image

### After

image (Updated as @ben wish.)

cf. [TortioseGit Settings Doc](https://tortoisegit.org/docs/tortoisegit/tgit-dug-settings.html) - the **Overlay Icon** section.


ben:
Hmm. Should we use the cover for this edition instead?

YueLinHo:
> Should we use the cover for this edition instead?

Yes.

Is this good enough?

image

or this?

image

or give me the font name of "Pro Git", I design another icon?

ben:
I like the top one the best, I think.

YueLinHo:
Updated the PR.

YueLinHo:
Thanks @ben. ^____^

YueLinHo:
Fixed #541

PR #544, opened 2016-02-25
Fix cmdline in credentials.asc

toolness:
I tried the following example from the book and it just displayed the help for `git config`:

```
git config --global credential.helper store --file ~/.my-credentials
```

After a bit of research I found the [git-credential-store(1) Manual Page](https://git-scm.com/docs/git-credential-store/1.7.12.1) which actually puts single quotes around the `store --file ~/.my-credentials`, which worked for me. I'm using git version 1.9.1, not sure if that makes a difference.

Note that there's another example at the bottom of this page, `git config --global credential.helper read-only --file /mnt/shared/creds`, which might also require single quotes around its last part to work, but I'm not sure, so I didn't change that.

### DIFF
--- a/book/07-git-tools/sections/credentials.asc
+++ b/book/07-git-tools/sections/credentials.asc
@@ -36,7 +36,7 @@ Here's an example of how you'd configure the ``store'' helper with a custom file
 
 [source,console]
 ----
-$ git config --global credential.helper store --file ~/.my-credentials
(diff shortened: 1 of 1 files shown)


ben:
Hey, nice catch. It definitely can't hurt with the `read-only` helper, why don't you go ahead and change that too.

toolness:
Sure thing, done!

ben:
:smile::+1:

PR #546, opened 2016-02-27
fixed to be more precise

KennethKinLum:
it really is a filename that ends with a `~`, instead of a file that ends with a `~`

### DIFF
--- a/book/02-git-basics/sections/recording-changes.asc
+++ b/book/02-git-basics/sections/recording-changes.asc
@@ -204,7 +204,7 @@ $ cat .gitignore
 ----
 
 The first line tells Git to ignore any files ending in ``.o'' or ``.a'' – object and archive files that may be the product of building your code.
-The second line tells Git to ignore all files that end with a tilde (`~`), which is used by many text editors such as Emacs to mark temporary files.
+The second line tells Git to ignore all files that end with a tilde (`~`) in the filename, which is used by many text editors such as Emacs to mark temporary files.
 You may also include a log, tmp, or pid directory; automatically generated documentation; and so on.
 Setting up a `.gitignore` file before you get going is generally a good idea so you don't accidentally commit files that you really don't want in your Git repository.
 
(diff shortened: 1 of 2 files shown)


book/02-git-basics/sections/recording-changes.asc
```diff
@@ -204,7 +204,7 @@ $ cat .gitignore
 ----
 
 The first line tells Git to ignore any files ending in ``.o'' or ``.a'' – object and archive files that may be the product of building your code.
-The second line tells Git to ignore all files that end with a tilde (`~`), which is used by many text editors such as Emacs to mark temporary files.
+The second line tells Git to ignore all files that end with a tilde (`~`) in the filename, which is used by many text editors such as Emacs to mark temporary files.
```
ben: What if we say "all files whose names end with a tilde"? I think that's slightly smoother.

KennethKinLum: ok... i changed and pushed...  you think 

If your preferred editor is Emacs or Vim, you know about files that end with a `~` or `.swp` in the filename.

will be

If your preferred editor is Emacs or Vim, you know about files whose names end with a `~` or `.swp`.

and 
The second line tells Git to ignore all files that end with a tilde (`~`) in the filename, which is used by many text editors such as Emacs to mark temporary files.

to 

The second line tells Git to ignore all files whose names end with a tilde (`~`), which is used by many text editors such as Emacs to mark temporary files.

may sound repetitive, but they are in different sections of the book.

ben: Yeah, those changes sound great.

KennethKinLum:
This case seems like can directly mention the "filenames":

If your preferred editor is Emacs or Vim, you know about filenames that end with a `~` or `.swp`.

book/02-git-basics/sections/recording-changes.asc
KennethKinLum: This case seems like can directly mention the "filenames":

If your preferred editor is Emacs or Vim, you know about filenames that end with a ~ or .swp.

ben:
I like it. Do you feel like this is ready?

KennethKinLum:
Sure, they look good to me.

Issue #550, opened 2016-03-04
Missing notice for a potentially dangerous command


nemoinho:
Chapter 9.2 "Migrating to git" has a paragraph which shows the use of `--no-metadata` to not include the metadata from svn. But there is no notice about the potential risk of this parameter if you may have to interact with the svn-repo again. The documentation says about this parameter that it is not recommended but in context of this chapter it seems to be a harmless option.

Documentation: (https://git-scm.com/docs/git-svn)
> --no-metadata
>
>    Set the noMetadata option in the [svn-remote] config. This option is not recommended, please read the svn.noMetadata section of this manpage before using this option.

And futher more the documentation describes why this paramater is risky: (https://git-scm.com/docs/git-svn#_config_file_only_options)
> This option is NOT recommended as it makes it difficult to track down old references to SVN revision numbers in existing documentation, bug reports and archives. If you plan to eventually migrate from SVN to Git and are certain about dropping SVN history, consider git-filter-branch[1] instead. filter-branch also allows reformatting of metadata for ease-of-reading and rewriting authorship info for non "svn.authorsFile" users.

YueLinHo:
Fixed with PR #555

nemoinho:
Thanks for the notice.

PR #555, opened 2016-03-09
Add recommendation note

nemoinho:
#550 Add note that it is not recommended to use --no-metadata because it may breaks
synchronisation between git and svn.


ben:
Hmm. Ordinarily I'd say this is good, but since this section is about doing a one-way conversion to Git, we may not actually want the metadata in the commit messages. Can we reword the warning to be something more like this?

> You can also tell `git svn` not to include the metadata that Subversion normally imports, by passing `--no-metadata` to the `clone` or `init` command (though if you want to keep the metadata, feel free to omit this parameter).

nemoinho:
Yes, this is also ok, but let me introduce one more word to your suggestion:
> You can also tell git svn not to include the metadata that Subversion normally imports, by passing --no-metadata to the clone or init command (though if you want to keep the _synchronisation_-metadata, feel free to omit this parameter).
So I changed the pull-request like that.

ben:
Nice. Thanks!

PR #567, opened 2016-03-16
Fix issue #312: /srv/git instead of /opt/git for server repositories

### LINKED ISSUE #312: /srv/git instead of /opt/git for server repositories
Section, "4.2 Git on the Server - Getting Git on a Server" uses /opt/git as the example location for putting repositories on a Linux server. Based on hier(7) it seems that /srv/git would be more consistent with filesystem conventions.

YueLinHo:
Fixes #312


ben:
:+1

YueLinHo:
Thank @ben :smiley:

Issue #573, opened 2016-03-21
json 1.8.1 won't build well on OS X with Ruby 2.3 (brew)


boltomli:
Installing json 1.8.1 with native extensions

Gem::Ext::BuildError: ERROR: Failed to build gem native extension.

    current directory: /usr/local/lib/ruby/gems/2.3.0/gems/json-1.8.1/ext/json/ext/generator
/usr/local/opt/ruby/bin/ruby -r ./siteconf20160321-22471-5lwbkr.rb extconf.rb
creating Makefile

current directory: /usr/local/lib/ruby/gems/2.3.0/gems/json-1.8.1/ext/json/ext/generator
make "DESTDIR=" clean

current directory: /usr/local/lib/ruby/gems/2.3.0/gems/json-1.8.1/ext/json/ext/generator
make "DESTDIR="
compiling generator.c
In file included from generator.c:1:
./../fbuffer/fbuffer.h:175:47: error: too few arguments provided to function-like macro invocation
    VALUE result = rb_str_new(FBUFFER_PAIR(fb));
                                              ^
/usr/local/Cellar/ruby/2.3.0/include/ruby-2.3.0/ruby/intern.h:797:9: note: macro 'rb_str_new' defined here
#define rb_str_new(str, len) __extension__ (    \
        ^
In file included from generator.c:1:
./../fbuffer/fbuffer.h:175:11: warning: incompatible pointer to integer conversion initializing 'VALUE' (aka 'unsigned long') with an expression of type 'VALUE (const char *, long)' [-Wint-conversion]
    VALUE result = rb_str_new(FBUFFER_PAIR(fb));
          ^        ~~~~~~~~~~
1 warning and 1 error generated.
make: *** [generator.o] Error 1

make failed, exit code 2

Gem files will remain installed in /usr/local/lib/ruby/gems/2.3.0/gems/json-1.8.1 for inspection.
Results logged to /usr/local/lib/ruby/gems/2.3.0/extensions/x86_64-darwin-15/2.3.0/json-1.8.1/gem_make.out

boltomli:
PR #574 to address this one.

ben:
Merged #574, should be fixed now.

PR #574, opened 2016-03-21
Update json to 1.8.3 and resolve issue #573

boltomli:
Make it bundle on OS X with Ruby 2.3 from brew. All files build well on my box.


ben:
:+1: Works on my machine!

PR #580, opened 2016-04-02
Fix issue #577: One command is `git mv`, instead of `mv`.

### LINKED ISSUE #577: One command is `git mv`, instead of `mv`.
Section 2.2 - Git Basics - Recording Changes to the Repository - [Moving Files](https://git-scm.com/book/en/v2/Git-Basics-Recording-Changes-to-the-Repository#Moving-Files)
book\02-git-basics\sections\recording-changes.asc **line 598**

```
The only real difference is that `mv` is one command instead of three
 – it's a convenience function.
```

should `mv` be `git mv`?
```
The only real difference is that `git mv` is one command instead of three
 – it's a convenience function.
```

YueLinHo:
Fixes #577


YueLinHo:
Thank @ben ^_^

PR #584, opened 2016-04-07
Fix git clone simplegit-progit https url

taichunmin:
I think the clone url with `.git` suffix is correct.


ben:
Both will work on most Git hosts that also have a web interface (BitBucket, Stash, etc.). Personally, I actually prefer the cleaner URLs.

Is this a point of ambiguity that we should call out somewhere?

taichunmin:
Thank you for telling me the reason. I will close the PR.

PR #592, opened 2016-04-11
fix jgit javadoc link

JonnyJD:
Unfortunately there is no generic javadoc link anymore.
I linked the latest stable version.
Old version docs are not removed so this should keep working.

The previous link leads to a project page and there is no link to the javadoc on that page.

### DIFF
--- a/book/B-embedding-git/sections/jgit.asc
+++ b/book/B-embedding-git/sections/jgit.asc
@@ -155,7 +155,7 @@ Many other commands are available through the Git class, including but not limit
 This is only a small sampling of JGit's full capabilities.
 If you're interested and want to learn more, here's where to look for information and inspiration:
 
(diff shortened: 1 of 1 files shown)


YueLinHo:
As time goes by the link is changed.
Perhaps using http://www.eclipse.org/jgit/documentation/, something like this:
```
The official JGit API documentation can be found at 
http://www.eclipse.org/jgit/documentation/[].
```

ben:
Yeah, if we can't point to an always-the-latest URL, we should drop the reader at a place where that can be found.

ben:
@JonnyJD, want to make the change we suggested and push a new commit to this branch?

JonnyJD:
Sorry, I have somewhat this PR. I added the commit as suggested by @YueLinHo

ben:
✨ Thanks!

PR #600, opened 2016-04-20
Harden retrieval of author list from Subversion

pks-t:
In order to get a list of authors from a Subversion repository,
we extract the authors from the output of `svn log --xml`. We
then try to filter out the author field by grepping for 'author'.
This produces incorrect results in the case where a commit's
message contains the word 'author'.

Fix the issue by invoking `svn log --xml --quiet` instead, which
suppresses the commit's message.


ben:
👍

PR #619, opened 2016-05-14
Update undoing.asc

Fornost461:
Corrected spelling mistake, according to the Wiktionary (when meaning “to falsify, as records”, the word is [wangle](https://en.wiktionary.org/wiki/wangle), not [wrangle](https://en.wiktionary.org/wiki/wrangle)).

### DIFF
--- a/book/02-git-basics/sections/undoing.asc
+++ b/book/02-git-basics/sections/undoing.asc
@@ -34,7 +34,7 @@ You end up with a single commit – the second commit replaces the results of th
 [[_unstaging]]
 ==== Unstaging a Staged File
 
-The next two sections demonstrate how to wrangle your staging area and working directory changes.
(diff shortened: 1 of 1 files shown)


ben:
Interesting! I've always understood "wrangle" in the cowboy sense, as in _to herd_ or _to manipulate_, but I guess that's a North-Americanism. It's probably better if we avoid regional idioms for an international audience, though; what if we used "work with" instead?

Fornost461:
Sure! I guess it was not a misspelling then, I had no idea that its sense “to herd” could be interpreted as “to manipulate.” I could hardly find a reference to this in online dictionaries, so a change to “work with” would greatly help non-native speakers like myself.

ben:
I had never heard it in the sense that apparently every dictionary says it means. 😀 Thanks!

Fornost461:
You’re welcome!

PR #620, opened 2016-05-15
linebreak deletion

Fornost461:
In the PDF, page 113, a linebreak is automatically inserted after the dashes within the option “--all”, which makes ambiguous whether a space should be inserted or not.
(I made two commits because I made a mistake in the first commit and I’m not yet at ease with git, but the two commits are meant as a single one.)


YueLinHo:
The screenshot from pdf for the issue:

pdfxcview_2016-05-16_09-05-58-d

ben:
Lovely.

PR #621, opened 2016-05-22
Update submodules.asc

HatiEth:
Changed `git clean -fdx` to `git clean -ffdx` to ensure the submodule from the other branch actually gets removed.

Just using `git clean -fdx` would result in git skipping the repository during removals.


ben:
😯 Nice catch!

PR #624, opened 2016-06-01
env-filter

galeksandrp:


ben:
So this definitely makes the code a bit more succinct, and I _like_ that, but does it make the book better for the reader? I'd argue that the original assumes less implicit knowledge. The code makes it clear that if you just do `git commit-tree "$@"` without changing the env vars, you'll get the original commit, but if you alter `GIT_AUTHOR_*` first, you'll get an altered history.

If this were a script, I'd accept this without question. But since it's a book, we have to prefer worse code that better aids understanding.

PR #633, opened 2016-08-04
Adapted to current git version

mkarg:
The on-disk image of a git repository has changed in recent versions of w.r.t to how svn-git stores remote branches and tags. As a consequence the bash commands told in this section of the book did not work anymore. My proposal solves this issue. I tested it today against 2.1.4, so I assume the change happened with v2.0 or v2.1 and should work with newer Git versions, too.


ben:
Beautiful. Thanks!

PR #636, opened 2016-08-04
Change message to use imperative present tense

mikecharles:
Directly above this example commit message it's stated that one should use the imperative present tense, so I think it's best to show that right away in the following example.

Note that I'm redoing this PR, since my last one (#222) was so long ago I thought this would be easier than updating/merging/rebasing.


ben:
👍

PR #637, opened 2016-08-04
Added examples how to change the directory

haripetrov:
It's very useful to have example for each OS how to change the directory
in Git (especially for Windows because the syntax is different from the
default syntax — C:/Users/your_repository).
Second try to contribute to this project.


ben:
👍

PR #646, opened 2016-08-29
Reference the 'push.recurseSubmodules' config in the section on pushing submodule changes

rjsenft:
I was looking into submodules and subtrees recently, and one of the reasons I decided to go with submodules was because there was a config option that forced safer behavior for pushing submodule changes.

I figured it'd be nice if the section mentioned the config; hopefully this will help someone else consider using submodules if they can enforce a team wide use of the config.

### DIFF
--- a/book/07-git-tools/sections/submodules.asc
+++ b/book/07-git-tools/sections/submodules.asc
@@ -544,6 +544,8 @@ To https://github.com/chaconinc/MainProject
 As you can see there, Git went into the DbConnector module and pushed it before pushing the main project.
 If that submodule push fails for some reason, the main project push will also fail.
 
+You can enable these behaviors for all future pushes with the `push.recurseSubmodules` config.
+
 ===== Merging Submodule Changes
 
 If you change a submodule reference at the same time as someone else, you may run into some problems.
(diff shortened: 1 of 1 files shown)


book/07-git-tools/sections/submodules.asc
```diff
@@ -544,6 +544,8 @@ To https://github.com/chaconinc/MainProject
 As you can see there, Git went into the DbConnector module and pushed it before pushing the main project.
 If that submodule push fails for some reason, the main project push will also fail.
 
+You can enable these behaviors for all future pushes with the `push.recurseSubmodules` config.
+
```
ben: It's a good addition, but let's not make it a separate paragraph. How about tacking something like this onto the last paragraph?

> You can make this behavior the default by doing `git config push.recurseSubmodules true`.

rjsenft:
I wanted it to be in its own paragraph since I wanted to put a bit of space from the section that was talking about the on-demand behavior. push.recurseSubmodules can accept both "check" and "on-demand", and I kinda failed properly imply that the command could be used to enforce either behavior.

I've tried splitting it out so that each section mentions the command in its own context with the value that it needs. What do you think?

ben:
Brilliant. Thanks!

PR #652, opened 2016-09-21
Removed Posh-Git GitHub for Windows reference

Velocet:
I installed some of the latest GitHub for Windows versions on Windows 10 and no PoSh-Git gets installed, so i think this information deprecated.


ben:
Lovely. Thanks!

PR #657, opened 2016-09-21
Replace the obscure term 'vendoring' with 'copying'

### LINKED ISSUE #656: Replace the term 'vendoring' with 'copying'
The verb 'vendoring' is a bit obscure and could probably be replaced with 'copying'.

crd:
Closes #656


ben:
Good point. Thanks!

PR #670, opened 2016-10-18
Import better-quality cover image

### LINKED ISSUE #463: Better cover image
The image quality of the cover image is quite poor; looks like it was converted from a low-quality JPEG. Can you upload a better version and maybe even include the original PSD or whatever the source file is?

ben:
Re-imported from Apress source TIFF image.

Fixes #463. /cc @huguenin


ben:
The new image is slightly smaller in pixels and larger in MB, but the quality difference is pretty striking:

image

Issue #547, opened 2016-03-01
git reset `--hard` note may confuse?


YueLinHo:
book\02-git-basics\sections\undoing.asc line 60 vs line 82

line 60
```
$ git reset HEAD CONTRIBUTING.md
```

line 82
```
While `git reset` _can_ be a dangerous command if you call it with `--hard`, 
in this instance the file in your working directory is not touched.
Calling `git reset` without an option is not dangerous - it only touches your staging area.
```

The `--hard` is not used in that example, then I was confused that. :P

Let's change it to ?

```
While `git reset` _can_ be a dangerous command if you call it with `--hard`, 
calling `git reset` without an option is not dangerous - it only touches your staging area.
So that in this instance the file in your working directory is not touched.
```

YueLinHo:
OK, my poor English. So close it.

PR #676, opened 2016-11-03
Update git reset --hard note for clarity

kausar007:
The note is confusing as the example before it does not use `git reset --hard`. This was mentioned in closed issue #547.

### DIFF
--- a/book/02-git-basics/sections/undoing.asc
+++ b/book/02-git-basics/sections/undoing.asc
@@ -79,7 +79,7 @@ The `CONTRIBUTING.md` file is modified but once again unstaged.
 
 [NOTE]
 =====
-While `git reset` _can_ be a dangerous command if you call it with `--hard`, in this instance the file in your working directory is not touched.
+While `git reset` _can_ be a dangerous command if you call it with `--hard` as it changes files in your working directory, however in the above example the file in your working directory is not touched.
 Calling `git reset` without an option is not dangerous - it only touches your staging area.
 =====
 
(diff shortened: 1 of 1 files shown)


book/02-git-basics/sections/undoing.asc
```diff
@@ -79,7 +79,7 @@ The `CONTRIBUTING.md` file is modified but once again unstaged.
 
 [NOTE]
 =====
-While `git reset` _can_ be a dangerous command if you call it with `--hard`, in this instance the file in your working directory is not touched.
+While `git reset` _can_ be a dangerous command if you call it with `--hard` as it changes files in your working directory, however in the above example the file in your working directory is not touched.
```
ben: I actually think this makes it a tad less readable. What about this:

```
While `git reset` _can_ be a dangerous command (especially if you call it with `--hard`), in this instance the file in your working directory is not touched, so it's relatively safe.
```

kausar007: I think when reading the whole sentence, it is possible that readers get confused about the usage of the phrase "in this instance" thinking that it is used for `git reset --hard` but the intention is the example before that. So using words like "in above example" or "in the instance above" would make it much clearer.

ben: OK, how about this:

```
It's true that `git reset` can be a dangerous command, especially if you provide the `--hard` flag.
However, in the scenario described above, the file in your working directory is not touched, so it's relatively safe.
```

ben:
✨ Thanks!

book/02-git-basics/sections/undoing.asc
kausar007: Perfect. Thanks. I have updated and pushed the changes.

PR #677, opened 2016-11-05
Add git remote remove command

kausar007:
Most of the `git remote` command names are verbs e.g add, show, rename etc. It's a good idea to add `git remote remove` as the first one. Easy to remember with other commands (add/remove). Hence I also updated the example.


ben:
Yeah, especially in this early chapter, it's probably better to use full words instead of unixy abbreviations. Thanks!

PR #678, opened 2016-11-05
8.2: Surround docx path with quotes

patrick96:
Paths and filenames containing spaces would not be diffed without the
quotes


ben:
😯 Nice catch, thanks!

PR #685, opened 2016-12-02
added escaped quotation marks for git commit's message.

aollier:
It is necessary to escape quotation marks in the git message, otherwise it ends the commit's message.

### DIFF
--- a/book/08-customizing-git/sections/attributes.asc
+++ b/book/08-customizing-git/sections/attributes.asc
@@ -329,9 +329,9 @@ The substitutions can include for example the commit message and any `git notes`
 [source,console]
 ----
 $ echo '$Format:Last commit: %h by %aN at %cd%n%+w(76,6,9)%B$' > LAST_COMMIT
-$ git commit -am 'export-subst uses git log's custom formatter
+$ git commit -am 'export-subst uses git log'\''s custom formatter
 
-git archive uses git log's `pretty=format:` processor
+git archive uses git log'\''s `pretty=format:` processor
 directly, and strips the surrounding `$Format:` and `$`
 markup from the output.
 '


book/08-customizing-git/sections/attributes.asc
```diff
@@ -329,9 +329,9 @@ The substitutions can include for example the commit message and any `git notes`
 [source,console]
 ----
 $ echo '$Format:Last commit: %h by %aN at %cd%n%+w(76,6,9)%B$' > LAST_COMMIT
-$ git commit -am 'export-subst uses git log's custom formatter
+$ git commit -am 'export-subst uses git log'\''s custom formatter
```
ben: I think you're right, but shouldn't this be:

```
$ git commit -am 'export-subst uses git log\'s custom formatter
```
?

jnavila: Hum, I have another guess:

    $ git commit -am 'export-subst uses git log\'s custom formatter'

with a quote at the end.

aollier: No, it does not work. You can try with a simple echo :
$ echo 'export-subst uses git log\'s custom formatter'
> ^C
$ echo 'export-subst uses git log'\''s custom formatter'
export-subst uses git log's custom formatter

With the first command, bash is waiting for rest of the command, which shows that using the back slash in the string does not do what you wanted.
With the second command, you get the expected result.

aollier: No, this does not work. You can try with a simple `echo`:
```
$ echo 'export-subst uses git log\'s custom formatter'
> ^C
$ echo 'export-subst uses git log'\''s custom formatter'
export-subst uses git log's custom formatter
```
With the first command, bash is waiting for the rest of the command.
With the second command, you get the expected result.

aollier: And Ben's proposition does not work better, it is not what you expect:
```
$ echo 'export-subst uses git log\'s custom formatter
export-subst uses git log\s custom formatter
```

jnavila: So please change your PR with the correct answer :smile:

aollier: The commit message is on several lines. This is the whole commit message as it will appear:
```
export-subst uses git log's custom formatter

git archive uses git log's `pretty=format:` processor
directly, and strips the surrounding `$Format:` and `$`
markup from the output.
```
So nothing need to be changed, right ?

jnavila: yup, my bad. OK for me.

lazarljubenovic:
Where can I look up _why_ does this work? I'm very confused why simply escaping the quote `\'` doesn't work. Whats the full story behind `'\''`?

jnavila:
If you have a bash at hand, you can try for yourself. when using a single quoted string, you cannot simply escape the inside quotes with a backslash. You have to
1.  leave the first part of the quoted string (first quote character)
2. escape the quote you want from the shell (\')
3. then reenter the quotation (last quote character).

Searching for more information, I saw that the original scheme could work if the whole quote is prefixed with a `$` character, like this (if it was only one line):

    echo $'export-subst uses git log\'s custom formatter'

PR #686, opened 2016-12-03
Bazaar

aollier:
I created two new sections that describe how to use Git as a Bazaar client and how to convert a Bazaar repository into a Git repository.

Thank you for reading carefully and validate my work if it is good for you.

### DIFF
--- a/book/09-git-and-other-scms/sections/client-bzr.asc
+++ b/book/09-git-and-other-scms/sections/client-bzr.asc
@@ -34,7 +34,7 @@ $ git clone bzr::bzr+ssh://developer@mybazaarserver:myproject myProject-Git
 $ cd myProject-Git
 ----
 
-At this point, your Git repository is created but it is not optimized at the point of view of its disk size.
+At this point, your Git repository is created but it is not compacted for optimal disk use.
 That's why you should also clean and compact your Git repository, especially if it is a big one:
 [source,console]
 ----
@@ -91,7 +91,7 @@ $ git pull --rebase origin
 
 ===== Push your work on the remote repository
 
-Because Bazaar supports very well merge commits, there will be no problem if you push a merge commit.
+Because Bazaar also has the concept of merge commits, there will be no problem if you push a merge commit.
 So you can work on a branch, merge the changes into `master` and push your work.
 Then, you create your branches, you test and commit your work as usual.
 You finally push your work to the Bazaar repository:
@@ -102,7 +102,8 @@ $ git push origin master
 
 ===== Caveats
 
-Limitations of the remote-helpers' framework apply. In particular, these commands don't work:
+Git's remote-helpers framework has some limitations that apply.
+In particular, these commands don't work:
 
 * git push origin :branch-to-delete (Bazaar can't accept ref deletions in this way.)
 * git push origin old:new (it will push 'old')
--- a/book/09-git-and-other-scms/sections/client-bzr.asc
+++ b/book/09-git-and-other-scms/sections/client-bzr.asc
@@ -1,0 +1,102 @@
+==== Git and Bazaar
+
+There are many projects that allow you to use Git as a Bazaar client.
+Here we'll use Felipe Contreras' project that you may find at https://github.com/felipec/git-remote-bzr[].
+To install it, you just have to download the file git-remote-bzr in a folder contained in your `$PATH`:
+[source,console]
+----
+$ wget https://raw.github.com/felipec/git-remote-bzr/master/git-remote-bzr -O ~/bin/git-remote-bzr
+$ chmod +x ~/bin/git-remote-bzr
+----
+
+You also need to have Bazaar installed.
+
+That's all!
+
+===== Create a Git repository from a Bazaar repository
+
+It is simple to use.
+It is enough to clone a Bazaar repository prefixing it by `bzr::`.
+If you worked in a local Bazaar repository and pushed to a remote one, it is better to clone the remote.
+That will make your work simpler; else you must go through your local repository and you may have problems when pushing to the remote if your Bazaar and Git repositories were not up-to-date.
+
+Let's suppose that you worked with a remote repository which is at address `bzr+ssh://developer@mybazaarserver:myproject`.
+Then you must clone it in the following way:
+[source,console]
+----
+$ git clone bzr::bzr+ssh://developer@mybazaarserver:myproject myProject-Git
+$ cd myProject-Git
+----
+
+You should also clean and compact the created Git repository, especially if it is a big one:
+[source,console]
+----
+$ git gc --aggressive
+----
+
+===== Bazaar branches
+
+Bazaar only allows you to clone branches, but a repository may contain several branches, and `git-remote-bzr` can clone both.
+For example, to clone a branch:
+[source,console]
+----
+$ git config remote-bzr.branches 'trunk, xwindow'
+----
+
+And to clone the whole repository:
+[source,console]
+----
+$ git clone bzr::bzr://bzr.savannah.gnu.org/emacs emacs
+----
+
+The second command clones all the branches contained in the emacs repository; nevertheless, it is possible to point out some branches:
+[source,console]
+----
+$ git config remote-bzr.branches 'trunk, xwindow'
+----
+
+Some remote repositories don't allow to list their branches, in which case you have to manually specify them, and even though you could specify the configuration in the cloning command, you may find this easier:
+
+[source,console]
+----
+$ git init emacs
+$ git remote add origin bzr::bzr://bzr.savannah.gnu.org/emacs
+$ git config remote-bzr.branches 'trunk, xwindow'
+$ git fetch
+----
+
+===== Ignore what is ignored with .bzrignore
+
+As the format of the .bzrignore file is completely compatible with .gitignore's one, and as you shouldn't make a .gitignore file in your repository, it is enough to make a symbolic link to .bzrignore so that the potential changes of .bzrignore are taken into account:
+[source,console]
+----
+$ ln -s .bzrignore .git/info/exclude
+----
+
+===== Fetch the changes of the remote repository
+
+To fetch the changes of the remote, you pull changes as usually, using Git commands.
+Supposing that your changes are on the `master` branch, you merge or rebase your work on the `origin/master` branch:
+[source,console]
+----
+$ git pull --rebase origin
+----
+
+===== Push your work on the remote repository
+
+Then, you create your branches, you test and commit your work as usual.
+You finally push your work to the Bazaar repository:
+[source,console]
+----
+$ git push origin master
+----
+
+===== Caveats
+
+There are some limitations while pushing to a Bazaar repository.
+Please refer to the README file of the project's Git repository (https://github.com/felipec/git-remote-bzr[]) for more information about this.
+
+===== Summary
+
+As Git and Bazaar are close, you can work in a Git repository and transparently push your work to a Bazaar repository.
+Be careful anyway to the limitations.
--- a/book/09-git-and-other-scms/sections/client-bzr.asc
+++ b/book/09-git-and-other-scms/sections/client-bzr.asc
@@ -1,5 +1,12 @@
 ==== Git and Bazaar
 
+Among the DVCS, another famous one is http://bazaar.canonical.com/[Bazaar].
+Bazaar is free and open source, and is part of the http://www.gnu.org/[GNU Project].
+It behaves very differently from Git.
+Sometimes, to do the same thing as with Git, you have to use a different keyword, and some keywords that are common don't have the same meaning.
+In particular, the branch management is very different and may cause confusion, especially when someone comes from Git's universe.
+Nevertheless, it is possible to work on a Bazaar repository from a Git one.
+
 There are many projects that allow you to use Git as a Bazaar client.
 Here we'll use Felipe Contreras' project that you may find at https://github.com/felipec/git-remote-bzr[].
 To install it, you just have to download the file git-remote-bzr in a folder contained in your `$PATH`:
@@ -10,15 +17,14 @@ $ chmod +x ~/bin/git-remote-bzr
 ----
 
 You also need to have Bazaar installed.
-
 That's all!
 
 ===== Create a Git repository from a Bazaar repository
 
 It is simple to use.
 It is enough to clone a Bazaar repository prefixing it by `bzr::`.
-If you worked in a local Bazaar repository and pushed to a remote one, it is better to clone the remote.
-That will make your work simpler; else you must go through your local repository and you may have problems when pushing to the remote if your Bazaar and Git repositories were not up-to-date.
+Since Git and Bazaar both do full clones to your machine, it's possible to attach a Git clone to your local Bazaar clone, but it isn't recommended.
+It's much easier to attach your Git clone directly to the same place your Bazaar clone is attached to ‒ the central repository.
 
 Let's suppose that you worked with a remote repository which is at address `bzr+ssh://developer@mybazaarserver:myproject`.
 Then you must clone it in the following way:
@@ -28,7 +34,8 @@ $ git clone bzr::bzr+ssh://developer@mybazaarserver:myproject myProject-Git
 $ cd myProject-Git
 ----
 
-You should also clean and compact the created Git repository, especially if it is a big one:
+At this point, your Git repository is created but it is not optimized at the point of view of its disk size.
+That's why you should also clean and compact your Git repository, especially if it is a big one:
 [source,console]
 ----
 $ git gc --aggressive
@@ -40,7 +47,7 @@ Bazaar only allows you to clone branches, but a repository may contain several b
 For example, to clone a branch:
 [source,console]
 ----
-$ git config remote-bzr.branches 'trunk, xwindow'
+$ git clone bzr::bzr://bzr.savannah.gnu.org/emacs/trunk emacs-trunk
 ----
 
 And to clone the whole repository:
@@ -67,7 +74,7 @@ $ git fetch
 
 ===== Ignore what is ignored with .bzrignore
 
-As the format of the .bzrignore file is completely compatible with .gitignore's one, and as you shouldn't make a .gitignore file in your repository, it is enough to make a symbolic link to .bzrignore so that the potential changes of .bzrignore are taken into account:
+As the format of the `.bzrignore` file is completely compatible with `.gitignore`'s one, and as you shouldn't make a `.gitignore` file in your repository, it is enough to make a symbolic link to `.bzrignore` so that the potential changes of `.bzrignore` are taken into account:
 [source,console]
 ----
 $ ln -s .bzrignore .git/info/exclude
@@ -84,6 +91,8 @@ $ git pull --rebase origin
 
 ===== Push your work on the remote repository
 
+Because Bazaar supports very well merge commits, there will be no problem if you push a merge commit.
+So you can work on a branch, merge the changes into `master` and push your work.
 Then, you create your branches, you test and commit your work as usual.
 You finally push your work to the Bazaar repository:
 [source,console]
@@ -93,10 +102,13 @@ $ git push origin master
 
 ===== Caveats
 
-There are some limitations while pushing to a Bazaar repository.
-Please refer to the README file of the project's Git repository (https://github.com/felipec/git-remote-bzr[]) for more information about this.
+Limitations of the remote-helpers' framework apply. In particular, these commands don't work:
+
+* git push origin :branch-to-delete (Bazaar can't accept ref deletions in this way.)
+* git push origin old:new (it will push 'old')
+* git push --dry-run origin branch (it will push)
 
 ===== Summary
 
-As Git and Bazaar are close, you can work in a Git repository and transparently push your work to a Bazaar repository.
-Be careful anyway to the limitations.
+Since Git's and Bazaar's models are similar, there isn't a lot of resistance when working across the boundary.
+As long as you watch out for the limitations, and are always aware that the remote repository isn't natively Git, you'll be fine.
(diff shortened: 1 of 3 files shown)


book/09-git-and-other-scms/sections/client-bzr.asc
```diff
@@ -0,0 +1,102 @@
+==== Git and Bazaar
```
ben: Instead of just jumping in, could you include a small introductory paragraph or two about Bazaar? Take a look at [the Mercurial section](https://github.com/aollier/progit2/blob/c85bf5f170343702d63b1ebefc897e516c8694f9/book/09-git-and-other-scms/sections/client-hg.asc) for the kind of thing I'm thinking of.

book/09-git-and-other-scms/sections/client-bzr.asc
```diff
@@ -0,0 +1,102 @@
...
+To install it, you just have to download the file git-remote-bzr in a folder contained in your `$PATH`:
+[source,console]
+----
+$ wget https://raw.github.com/felipec/git-remote-bzr/master/git-remote-bzr -O ~/bin/git-remote-bzr
+$ chmod +x ~/bin/git-remote-bzr
+----
+
+You also need to have Bazaar installed.
+
+That's all!
```
ben: Let's remove the blank line here, these two sentences fit together in one paragraph.

book/09-git-and-other-scms/sections/client-bzr.asc
```diff
@@ -0,0 +1,102 @@
...
+
+===== Create a Git repository from a Bazaar repository
+
+It is simple to use.
+It is enough to clone a Bazaar repository prefixing it by `bzr::`.
+If you worked in a local Bazaar repository and pushed to a remote one, it is better to clone the remote.
+That will make your work simpler; else you must go through your local repository and you may have problems when pushing to the remote if your Bazaar and Git repositories were not up-to-date.
```
ben: Let's rephrase a bit:

> Since Git and Bazaar both do full clones to your machine, it's possible to attach a Git clone to your local Bazaar clone, but this isn't recommended. It's much easier to attach your Git clone directly to the same place your Bazaar clone is attached to - the central repository.

…or something like that. Does that make sense?

book/09-git-and-other-scms/sections/client-bzr.asc
```diff
@@ -0,0 +1,102 @@
...
+
+Let's suppose that you worked with a remote repository which is at address `bzr+ssh://developer@mybazaarserver:myproject`.
+Then you must clone it in the following way:
+[source,console]
+----
+$ git clone bzr::bzr+ssh://developer@mybazaarserver:myproject myProject-Git
+$ cd myProject-Git
+----
+
+You should also clean and compact the created Git repository, especially if it is a big one:
```
ben: Can you give an idea of why? Git doesn't require a manual `gc` after a clone, why does `git-remote-bzr`?

book/09-git-and-other-scms/sections/client-bzr.asc
```diff
@@ -0,0 +1,102 @@
...
+[source,console]
+----
+$ git config remote-bzr.branches 'trunk, xwindow'
+----
+
+And to clone the whole repository:
+[source,console]
+----
+$ git clone bzr::bzr://bzr.savannah.gnu.org/emacs emacs
+----
```
ben: Will this command fail on a repository that doesn't list its branches? If so, it might be better to say:

> Cloning an entire repository is easy, just go like so:
> ```
> $ git clone bzr::bzr://bzr.savannah.gnu.org/emacs emacs
> ```
> However, some Bazaar repositories don't provide a listing of their branches, so you'll have to ask for them by name. This is managed using the `remote-bzr.branches` configuration variable:
> ```
> $ git init
> […]
> ```

book/09-git-and-other-scms/sections/client-bzr.asc
```diff
@@ -0,0 +1,102 @@
...
+$ git fetch
+----
+
+===== Ignore what is ignored with .bzrignore
+
+As the format of the .bzrignore file is completely compatible with .gitignore's one, and as you shouldn't make a .gitignore file in your repository, it is enough to make a symbolic link to .bzrignore so that the potential changes of .bzrignore are taken into account:
+[source,console]
+----
+$ ln -s .bzrignore .git/info/exclude
+----
```
ben: Will this result in the `.gitignore` file being listed as a "new file" when I type `git status`?

book/09-git-and-other-scms/sections/client-bzr.asc
```diff
@@ -0,0 +1,102 @@
...
+
+===== Push your work on the remote repository
+
+Then, you create your branches, you test and commit your work as usual.
+You finally push your work to the Bazaar repository:
+[source,console]
+----
+$ git push origin master
+----
```
ben: What happens if I push a merge commit? Should I prefer pushing commits directly to `master` or is it preferred to work in a topic branch with a recorded merge?

book/09-git-and-other-scms/sections/client-bzr.asc
```diff
@@ -0,0 +1,102 @@
...
+
+===== Caveats
+
+There are some limitations while pushing to a Bazaar repository.
+Please refer to the README file of the project's Git repository (https://github.com/felipec/git-remote-bzr[]) for more information about this.
+
+===== Summary
+
+As Git and Bazaar are close, you can work in a Git repository and transparently push your work to a Bazaar repository.
+Be careful anyway to the limitations.
```
ben: The English here is a bit stiff, let's try this:

> Since Git's and Bazaar's models are similar, there isn't a lot of resistance when working across the boundary. As long as you watch out for the limitations, and are always aware that the remote repository isn't natively Git, you'll be fine.

book/09-git-and-other-scms/sections/client-bzr.asc
```diff
@@ -0,0 +1,102 @@
...
+You finally push your work to the Bazaar repository:
+[source,console]
+----
+$ git push origin master
+----
+
+===== Caveats
+
+There are some limitations while pushing to a Bazaar repository.
+Please refer to the README file of the project's Git repository (https://github.com/felipec/git-remote-bzr[]) for more information about this.
```
ben: Let's go ahead and summarize those limitations here; there aren't that many, and sending the user completely out of their book-reading flow just to read a list of 4 things feels clunky.

book/09-git-and-other-scms/sections/import-bzr.asc
```diff
@@ -0,0 +1,154 @@
+==== Bazaar
+(((Bazaar)))(((Importing, from Bazaar)))
+
+Bazaar is a distributed source management content like Git.
+It is very easy to convert a Bazaar repository into a Git repository thanks to Bazaar's plugin system.
+To convert a Bazaar repository into a Git repository, you have first to import the bzr-fastimport plugin.
+Here is the whole procedure if you have not yet imported any plugin:
```
ben: Let's rephrase:

> Bazaar is a DVCS tool much like Git, and as a result it's pretty straightforward to convert a Bazaar repository into a Git one. To accomplish this, you'll need to import the `bzr-fastimport` plugin:
> _(code snippet)_

book/09-git-and-other-scms/sections/import-bzr.asc
```diff
@@ -0,0 +1,154 @@
...
+----
+
+This message is displayed because the fastimport python module that the `bzr fast-export` command needs is missing.
+Install it from your package manager if you are under GNU/Linux or download it at address https://pypi.python.org/pypi/fastimport/[] if you are under Windows.
+For example, with Fedora, you would do the following:
+
+[source,console]
+----
+$ sudo dnf install python-fastimport
+----
```
ben: This belongs in the section where we installed the plugin. Say something like

> To make sure this works, try this:
> ```
> $ [something that works if everything is installed correctly]
> ```
> If you get an error, you're probably missing a Python module. To correct this, do this:
> ```
> $ pip install fastimport
> ```

(Prefer a command that works almost anywhere the user could have Bazaar installed. Pip comes with most Python installations these days.)

book/09-git-and-other-scms/sections/import-bzr.asc
```diff
@@ -0,0 +1,154 @@
...
+$ mkdir --parents ~/.bazaar/plugins/bzr # creates the necessary folders for the plugins
+$ cd ~/.bazaar/plugins/bzr
+$ bzr branch lp:bzr-fastimport fastimport # imports the fastimport plugin
+$ cd fastimport
+$ sudo python setup.py install --record=files.txt
+----
+
+There you go!
+You've just installed the fastimport plugin.
```
ben: Instead of this, warn the reader that their use case is going to fall into one of two categories:

> Now you're ready to convert your repository. This procedure is slightly different depending on whether your Bazaar repo has one branch or many.

book/09-git-and-other-scms/sections/import-bzr.asc
```diff
@@ -0,0 +1,154 @@
...
+----
+$ sudo dnf install python-fastimport
+----
+
+Now you can export your Bazaar repository to the Git repository.
+
+Depending on the size of the project, your Git repository is built in a lapse from a few seconds to a few minutes.
+
+At this point, if you type `git status`, you'll see that the tracked files are now marked as removed for the next commit.
+Here is an example:
```
ben: Remove this sentence, it's obvious that it's an example.

book/09-git-and-other-scms/sections/import-bzr.asc
```diff
@@ -0,0 +1,154 @@
...
+
+Untracked files:
+  (use "git add <fichier>..." to include in what will be validated)
+
+        .bzr/
+        .bzrignore
+        file.txt
+----
+
+You can restore the repository in a correct state very simply with the `git reset` command:
```
ben: Tell the user why. It's because the `.git` folder has been generated, but the working directory isn't populated, like with most fast-imports. But they'll want to know so they know why they're typing the next thing.

book/09-git-and-other-scms/sections/import-bzr.asc
```diff
@@ -0,0 +1,154 @@
...
+
+You can also import a Bazaar repository that contains branches.
+Let us suppose that you have two branches: one represents the main branch (myProject.trunk), the other one is the working branch (myProject.work).
+[source,console]
+----
+$ ls
+myProject.trunk myProject.work
+----
+
+Create the Git repository and go into it:
```
ben: Use `cd` instead of "go".

book/09-git-and-other-scms/sections/client-bzr.asc
jnavila: No, that's why we don't want to link the `.bzrignore` to the file hidden in the .git directory. This way, only the `.bzrignore` file is managed and git behaves accordingly.

ben: Oh right, I guess I should read the actual command line. 😊

book/09-git-and-other-scms/sections/client-bzr.asc
aollier: Ok.

book/09-git-and-other-scms/sections/client-bzr.asc
aollier: I don't know why, this is recommended by Felipe Contrera himself in the section `Notes`.
I just copied his recommandation.

book/09-git-and-other-scms/sections/client-bzr.asc
aollier: Here is the content of the section `Caveats` :
> Limitations of the remote-helpers' framework apply. In particular, these commands don’t work:
> * git push origin :branch-to-delete
> * git push origin old:new (it will push 'old') (patches available)
> * git push --dry-run origin branch (it will push) (patches available)

But I found nowhere the available patches he is talking about, that's why I did not want to talk about something I don't understand very well.

book/09-git-and-other-scms/sections/import-bzr.asc
aollier: No, the bzr-fastimport module is mandatory for the two cases, so this part is common.
That's why it is just after this I subdivided my section into two parts.

book/09-git-and-other-scms/sections/import-bzr.asc
aollier: The working directory is actually populated because we are in the bzr repository which is populated.
I think the reason is that the staging area is not yet set, so `git status` does not display the correct state.
`git reset HEAD .` sets the index, that's why the next `git status` is correct.

book/09-git-and-other-scms/sections/import-bzr.asc
aollier: Why ? Isn't it clear with the verb to go ? Even if it was not the case, the command is given three lines later.

book/09-git-and-other-scms/sections/client-bzr.asc
aollier: I sent two emails to @felipec on this subject but he never answered.

book/09-git-and-other-scms/sections/client-bzr.asc
ben: Maybe try it out, and check the output of `git gc` to see if it did anything. My expectation is that it doesn't do much, because fast-import has gotten smarter over the years. 

Bottom line: if there's a real need to run this command, we should tell why. If there isn't, we should remove the recommendation.

book/09-git-and-other-scms/sections/client-bzr.asc
ben: I think it's safe to just note the limitations that are stated, and assume the patches don't exist. Readers of this book shouldn't be expected to track down and apply patches for remote helper.

book/09-git-and-other-scms/sections/import-bzr.asc
ben: No, I mean replace the two sentences you have above with something that leads into the next two sections.

book/09-git-and-other-scms/sections/import-bzr.asc
ben: Ah, I see. Well, tell the reader that, then:

> At this point, the `.git` directory and your working tree are all set up, but Git's index isn't synchronized with the working tree. Type this to fix that: […]

book/09-git-and-other-scms/sections/import-bzr.asc
ben: On a technical level, using "cd" as a verb is more correct. The user isn't physically "going" anywhere.

On a consistency level, I don't think we use "go" in this way in any other part of the book, so it would fit better to use "cd."

On an aesthetic level, it just seems to read better, at least to me.

book/09-git-and-other-scms/sections/client-bzr.asc
aollier: I did it several times and I saw that Git does some things. It spends some time to compress the `.git` directory so I think this is not superfluous.
If I have to add a reason, I would say something like this:
> At this point your git repository is created but it is not optimized at the point of view of its disk size.
> That's why you should also clean [...]

Do you agree with this formulation ?

book/09-git-and-other-scms/sections/client-bzr.asc
aollier: Bazaar supports very well merge commits. You can work in a branch, merge it in the main branch and push it, exactly like with Git.
Bazaar supports also rebasing thanks to a plugin.
So you can do as you want. There is no technical limitation, it is more linked with a policy of the project.

book/09-git-and-other-scms/sections/client-bzr.asc
aollier: Ok so should I remove the parts mentionning patches ?

book/09-git-and-other-scms/sections/import-bzr.asc
aollier: I think it is better to add a sentence to introduce it, because I think it is important to mention that at this point, the plugin is installed. Do you agree ?

book/09-git-and-other-scms/sections/client-bzr.asc
ben: That's the hard part about putting it in a book. We have to do the subject justice and be a useful resource, which sometimes means writing more documentation than the project author has.

book/09-git-and-other-scms/sections/client-bzr.asc
ben: Sounds good.

book/09-git-and-other-scms/sections/client-bzr.asc
ben: Go ahead and mention that then. Given that both tools have a "merge commit" feature, the reader will wonder whether that transfers from one to the other.

book/09-git-and-other-scms/sections/client-bzr.asc
ben: Yeah, and rephrase the limitation a bit.

book/09-git-and-other-scms/sections/import-bzr.asc
ben: The goal of this bit of text is to (a) let the user know they've reached a stopping point if they're following along, and (b) set them up to understand what's coming next. So yes, say they've done it, and what's coming in the next section.

book/09-git-and-other-scms/sections/client-bzr.asc
jnavila: > that's why we *want* 
My bad, my explanation was the inverse of what I meant. Glad you caught the meaning despite that. :1st_place_medal:

book/09-git-and-other-scms/sections/client-bzr.asc
aollier: Yes that makes sense, that says exactly the same thing, but said differently.
So I'll use your suggestion.

book/09-git-and-other-scms/sections/client-bzr.asc
aollier: What do you think about this introduction ?
> Among the DVCS, another famous one is [Bazaar](http://bazaar.canonical.com/en/).
> Bazaar is free and open source, and is part of the [GNU Project](http://www.gnu.org/).
> It behaves very differently from Git.
> Sometimes, to do the same thing as with Git, you have to use a different keyword, and some keywords that are common don't have the same meaning.
> The branch management is very different and may cause confusion, especially when someone comes from Git's universe.
> But strangely and fortunately, it is simple to switch from Bazaar to Git.

I made an adaption of the URLs for the format for markdown but I'll use the appropriate one for asciidoctor.

ben: Perfect!

book/09-git-and-other-scms/sections/client-bzr.asc
aollier: The command
```console
$ git clone bzr::bzr://bzr.savannah.gnu.org/emacs emacs
```
will not fail because there is at least the main branch that will be available.
Bazaar has a strange way to manage branches. There are several types of branches and we can't study the different cases. It is not the good place to do that.

With repositories that do not allow to list the branches, I don't know if specifying the branches with the commands `git init`, `git remote add ...`, `git config remote-bzr.branches ...` and finally `git fetch` works. I don't know how this case is possible so I can't test it.
So I think the simplest is to let it as it is.
Unfortunately, I think the only thing we can do is to trust the author.

Maybe if some day someone runs into this case and realizes that something is wrong, he or she will be able to correct this section or add some details. Don't you agree ?

book/09-git-and-other-scms/sections/client-bzr.asc
```diff
@@ -0,0 +1,102 @@
...
+$ git gc --aggressive
+----
+
+===== Bazaar branches
+
+Bazaar only allows you to clone branches, but a repository may contain several branches, and `git-remote-bzr` can clone both.
+For example, to clone a branch:
+[source,console]
+----
+$ git config remote-bzr.branches 'trunk, xwindow'
```
aollier:
```console
$ git clone bzr::bzr://bzr.savannah.gnu.org/emacs emacs
```

The final "emacs" is optionnal since the new repository will have the name of the last part of the URL.

book/09-git-and-other-scms/sections/import-bzr.asc
aollier: I'll remove the example, because it is not necessary. Do you agree ?

book/09-git-and-other-scms/sections/client-bzr.asc
ben: Okay, I'm fine with that. What you've got here seems to be correct.

book/09-git-and-other-scms/sections/import-bzr.asc
ben: Actually, I think the example that follows is valuable. It's the first step in showing the reader what their Git repo will look like after they've converted it.

book/09-git-and-other-scms/sections/client-bzr.asc
aollier: Any suggestion for rephrasing ?

book/09-git-and-other-scms/sections/import-bzr.asc
aollier: I added some instructions to install bzr-fastimport since I realized that it is available in the Fedora and Debian repositories. I first suggest to install it using the usual command, quoting some of the main Linux distributions (I put the examples using apt-get, yum and dnf).

book/09-git-and-other-scms/sections/client-bzr.asc
ben: Something like:

> `git push origin :branch-to-delete` (Bazaar can't accept ref deletions in this way.)

ben:
Looking good. Do you feel like it's ready for a final read-through?

aollier:
Yes I think so.

ben:
Let's do it then!

PR #691, opened 2016-12-15
Corrected a little typo in a section

sivaraam:
As far as I could think of the word _based_ would fit better than the word _basic_ in that line I changed in the PR. Correct me, if i'm wrong. 

P.S. : Thanks for such a good book on Git

### DIFF
--- a/book/04-git-server/sections/protocols.asc
+++ b/book/04-git-server/sections/protocols.asc
@@ -75,7 +75,7 @@ We'll cover the newer ``smart'' HTTP protocol first.
 ===== Smart HTTP
 
 (((protocols, smart HTTP)))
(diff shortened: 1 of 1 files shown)


ben:
Hmm. The term "basic authentication" has a [technical definition](https://tools.ietf.org/html/rfc2617) in the context of HTTP, but I get how that sentence doesn't read particularly well. What if we went with something like this instead:

> […] since you can use things like username/password authentication rather than having to set up SSH keys.

Oh, and you're welcome! I'm glad you like it, thanks for your contribution!

sivaraam:
@ben That would do fine too, sir. I'll make the change.

ben:
✨  Thanks!

sivaraam:
You're welcome!

PR #699, opened 2016-12-24
Correct wrong statement on whitespace config

shengbinmeng:
Well, this is actually a followup of https://github.com/progit/progit2/pull/698 .

As mentioned in that PR, I found that "leaving them out of the setting string" doesn't disable the settings, but will only keep the default value unchanged. Commits in this PR try to reflect that.

### DIFF
--- a/book/08-customizing-git/sections/config.asc
+++ b/book/08-customizing-git/sections/config.asc
@@ -421,7 +421,7 @@ The three that are turned on by default are `blank-at-eol`, which looks for spac
 The three that are disabled by default but can be turned on are `indent-with-non-tab`, which looks for lines that begin with spaces instead of tabs (and is controlled by the `tabwidth` option); `tab-in-indent`, which watches for tabs in the indentation portion of a line; and `cr-at-eol`, which tells Git that carriage returns at the end of lines are OK.
 
 You can tell Git which of these you want enabled by setting `core.whitespace` to the values you want on or off, separated by commas.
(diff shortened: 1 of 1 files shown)


book/08-customizing-git/sections/config.asc
```diff
@@ -421,13 +421,21 @@ The three that are turned on by default are `blank-at-eol`, which looks for spac
 The three that are disabled by default but can be turned on are `indent-with-non-tab`, which looks for lines that begin with spaces instead of tabs (and is controlled by the `tabwidth` option); `tab-in-indent`, which watches for tabs in the indentation portion of a line; and `cr-at-eol`, which tells Git that carriage returns at the end of lines are OK.
 
 You can tell Git which of these you want enabled by setting `core.whitespace` to the values you want on or off, separated by commas.
-You can disable settings by either leaving them out of the setting string or prepending a `-` in front of the value.
-For example, if you want all but `cr-at-eol` to be set, you can do this (with `trailing-space` being a short-hand to cover both `blank-at-eol` and `blank-at-eof`):
+You can disable settings by prepending a `-` in front of the value, or use default value of settings by leaving them out of the setting string.
```
ben: This is good, just one correction:

> You can disable an option by prepending a `-` in front of it's name, or use the default value by leaving it out of the string entirely.

shengbinmeng: Sure thing! Thank you.

PR #702, opened 2016-12-30
Fixes #312

### LINKED ISSUE #312: /srv/git instead of /opt/git for server repositories

johnhar:
/srv/git instead of /opt/git for server repositories. Fixed in all of
chapter 4


ben:
Brilliant, thanks!

PR #706, opened 2017-01-15
Fix link

pascalberger:
Fix wrong link introduced by #705.

Sorry, so used to Markdown syntax :)


ben:
Hah, whoops! I missed that too. Thanks!

PR #710, opened 2017-01-19
Added a few changes to the "GitHub" chapter

sivaraam:
A few suggestions are required for the following change,

- In the last code sample found in the _Commenting on an Issue_ subsection of the _Scripting GitHub_ section (just above figure 6-54) has a line like this,

```
    $ curl -H "Content-Type: application/json" \
     -H "Authorization: token TOKEN" \
      ....
```
The second line seems to need a token generated using GItHub in place of which the word TOKEN is placed. Wouldn't it be nice to add a note about that in that section?

### DIFF
--- a/book/06-github/1-github.asc
+++ b/book/06-github/1-github.asc
@@ -2,7 +2,7 @@
 == GitHub
 
 (((GitHub)))
-GitHub is one of the largest host for Git repositories, and is the central point of collaboration for millions of developers and projects.
+GitHub is the single largest host for Git repositories, and is the central point of collaboration for millions of developers and projects.
 A large percentage of all Git repositories are hosted on GitHub, and many open-source projects use it for Git hosting, issue tracking, code review, and other things.
 So while it's not a direct part of the Git open source project, there's a good chance that you'll want or need to interact with GitHub at some point while using Git professionally.
 
--- a/book/06-github/1-github.asc
+++ b/book/06-github/1-github.asc
@@ -2,7 +2,7 @@
 == GitHub
 
 (((GitHub)))
-GitHub is the single largest host for Git repositories, and is the central point of collaboration for millions of developers and projects.
+GitHub is one of the largest host for Git repositories, and is the central point of collaboration for millions of developers and projects.
 A large percentage of all Git repositories are hosted on GitHub, and many open-source projects use it for Git hosting, issue tracking, code review, and other things.
 So while it's not a direct part of the Git open source project, there's a good chance that you'll want or need to interact with GitHub at some point while using Git professionally.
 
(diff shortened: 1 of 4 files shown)


book/06-github/sections/2-contributing.asc
```diff
@@ -407,12 +406,12 @@ There are actually quite a number of web services that make use of emoji charact
...
+===== Images
 
 This isn't technically GitHub Flavored Markdown, but it is incredibly useful. In addition to adding Markdown image links to comments, which can be difficult to find and embed URLs for, GitHub allows you to drag and drop images into text areas to embed them.
 
 [[_md_drag]]
 .Drag and drop images to upload them and auto-embed them.
 image::images/markdown-08-drag-drop.png[Drag and drop images]
 
-If you look back at <<_pr_references>>, you can see a small ``Parsed as Markdown'' hint above the text area. Clicking on that will give you a full cheat sheet of everything you can do with Markdown on GitHub.
+If you at <<_md_drag>>, you can see a small ``Parsed as Markdown'' hint above the text area. Clicking on that will give you a full cheat sheet of everything you can do with Markdown on GitHub.
```
aollier: A verb is missing in the first part of this sentence, the one you removed (look back).

book/06-github/1-github.asc
```diff
@@ -2,7 +2,7 @@
 == GitHub
 
 (((GitHub)))
-GitHub is the single largest host for Git repositories, and is the central point of collaboration for millions of developers and projects.
+GitHub is one of the largest host for Git repositories, and is the central point of collaboration for millions of developers and projects.
```
ben: This change is incorrect. As of right now, GitHub is still the largest single Git host. We only included a chapter on GitHub because the average reader will most likely be using it.

sivaraam:
I have made he changes as requested. Any suggestions regarding my initial comment ?

sivaraam:
@ben I guess you misunderstood my question. I asked suggestion regarding the following,

>In the last code sample found in the Commenting on an Issue subsection of the Scripting GitHub section (just above figure 6-54) has a line like this,

    $ curl -H "Content-Type: application/json" \
     -H "Authorization: token TOKEN" \
      ....

>The second line seems to need a token generated using GItHub in place of which the word TOKEN is placed. Wouldn't it be nice to add a note about that in that section?

ben:
Oh yeah, sorry about that. I don't think it's necessary, since the [section that line is in](https://git-scm.com/book/en/v2/GitHub-Scripting-GitHub#_commenting_on_an_issue) spends 4 entire paragraphs walking through generating a token, and says this right before the code snippet:

> do an HTTP POST request to `repos/<user>/<repo>/issues/<num>/comments` with the token we just generated as an Authorization header.

sivaraam:
👍

PR #712, opened 2017-01-21
Remove references to Atlas.

jnavila:
The build is no longer performed by Atlas. We are on our own for publishing.

@ben do we need more on this?

### DIFF
--- a/README.asc
+++ b/README.asc
@@ -8,7 +8,6 @@ Like the first edition, the second edition of Pro Git is open source under a Cre
 
 A couple of things have changed since open sourcing the first edition.
 For one, we've moved from Markdown to the amazing Asciidoc format for the text of the book.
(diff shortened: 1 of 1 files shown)


README.asc
```diff
@@ -20,7 +19,7 @@ There are two ways to generate e-book content from this source code.
 The easiest way is simply to let us do it.
 A robot is standing by to look for new work on the main branch and automatically build it for everyone.
 
-You can find the current builds on http://git-scm.com/book[] and more information about the builds available at https://progit.org[].
+You can find the current builds on http://git-scm.com/book[] .
```
ben: There's an extraneous space here.

jnavila: Good catch!

jnavila:
Unfortunately, we only update the website, there's no longer any pdf or ebook. There's a discussion going on with @peff about this issue.

ben:
Fair enough, we'll figure that out separately. Thanks!

PR #713, opened 2017-01-23
The repo needs a reset after the import from bzr

jnavila:
just for a quick review before inclusion.


ben:
👍, merging. /cc @aollier

Issue #718, opened 2017-02-08
dependencies for installing documentation


flip111:
http://git-scm.com/book/en/v2/Getting-Started-Installing-Git#Installing-from-Source

this section does not mention you need `asciicode` -> `xmlto` -> `getopt`

original issue from: https://github.com/git/git-scm.com/issues/493

ben:
I'd _love_ for these directions to be more complete and correct. Care to submit a PR? The file you want to modify is [this one](https://github.com/progit/progit2/blob/master/book/01-introduction/sections/installing.asc).

flip111:
ok https://github.com/progit/progit2/pull/719

PR #719, opened 2017-02-08
Update installing.asc

flip111:
add getopt to dependencies


ben:
Nice, thanks! For future reference, you could have just put "fixes #718" in the description of this PR, and the issue would be automatically closed when the PR was merged.

PR #724, opened 2017-02-17
2-contributing.asc: sed command fix/type

apiemont:
in Chapt. 6 Sect. 2, the following command seems to have a typo, causing the error below: 

>$ sed -i '' 's/1000/3000/'  blink.ino
>sed: can't read s/1000/3000/: No such file or directory

### DIFF
--- a/book/06-github/sections/2-contributing.asc
+++ b/book/06-github/sections/2-contributing.asc
@@ -72,7 +72,8 @@ $ git checkout -b slow-blink <2>
 Switched to a new branch 'slow-blink'
 
 $ sed -i '' 's/1000/3000/' blink.ino <3> (MacOSX)
-$ sed -i 's/1000/3000/' blink.ino <3> (Linux)
+# If you're on a Linux system, do this instead:
+# $ sed -i 's/1000/3000/' blink.ino
 
 $ git diff --word-diff <4>
 diff --git a/blink.ino b/blink.ino
--- a/book/06-github/sections/2-contributing.asc
+++ b/book/06-github/sections/2-contributing.asc
@@ -71,7 +71,7 @@ $ cd blink
 $ git checkout -b slow-blink <2>
 Switched to a new branch 'slow-blink'
 
-$ sed -i '' 's/1000/3000/' blink.ino <3>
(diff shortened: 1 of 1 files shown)


devwebcl:
I think it's correct. I didn't have any issues using that sed sentence.
Actually It changed 1000 by 3000:

```
$ git diff
diff --git a/blink.ino b/blink.ino
index 15b9911..a6cc5a5 100644
--- a/blink.ino
+++ b/blink.ino
@@ -18,7 +18,7 @@ void setup() {
 // the loop routine runs over and over again forever:
 void loop() {
   digitalWrite(led, HIGH);   // turn the LED on (HIGH is the voltage level)
-  delay(1000);               // wait for a second
+  delay(3000);               // wait for a second
   digitalWrite(led, LOW);    // turn the LED off by making the voltage LOW
-  delay(1000);               // wait for a second
+  delay(3000);               // wait for a second
 }

```

apiemont:
Maybe the reason is that you are using BSD sed (on OSX)? I used GNU sed on linux and I get the error I reported.

apiemont:
made new commit to indicate corect command based on OS (63bba9d0a2b98d412d976787fd4a92fd42665db8)

book/06-github/sections/2-contributing.asc
```diff
@@ -71,7 +71,8 @@ $ cd blink
 $ git checkout -b slow-blink <2>
 Switched to a new branch 'slow-blink'
 
-$ sed -i '' 's/1000/3000/' blink.ino <3>
+$ sed -i '' 's/1000/3000/' blink.ino <3> (MacOSX)
+$ sed -i 's/1000/3000/' blink.ino <3> (Linux)
```
ben: Let's format this a bit differently:

```shell
$ sed -i '' 's/1000/3000/' blink.ino <3>
# If you're on a Linux system, do this instead:
# $ sed -i 's/1000/3000/' blink.ino
```

That way the "3" bullet only shows up once.

apiemont: sounds good to me, changed accordingly in 88d99d440d6c1d0d9aa74d3609d3227ddb400b5e

PR #726, opened 2017-02-21
Fix figure title

### LINKED ISSUE #725: Section 10.2: Git Internals - Git Objects, Figure 3 is not "All the objects in your Git directory."
References here:

https://github.com/progit/progit2/blob/master/book/10-git-internals/sections/objects.asc

which points to https://github.com/progit/progit2/blob/master/book/10-git-internals/images/data-model-3.png

is depiction of all the objects but missing the blob

`d670460b4b4aece5915caf5c68d12f560a9fe3e4`

Basically the object listing by `find .git/objects -type f` shows 10 objects, but the figure shows 9. The missing one is the orphan one - not pointed by any other objects. Therefore the caption isn't really "All the objects in your Git directory." :)

saadlu:
The orphan blob d670460b4b4aece5915caf5c68d12f560a9fe3e4 is not in the
figure. So the figure is not of "All the objects" but rather all the
reachable ones.

Fixes #725.

### DIFF
--- a/book/10-git-internals/sections/objects.asc
+++ b/book/10-git-internals/sections/objects.asc
@@ -329,7 +329,7 @@ $ find .git/objects -type f
 If you follow all the internal pointers, you get an object graph something like this:
 
 .All the reachable objects in your Git directory.
-image::images/data-model-3.png[All the objects in your Git directory.]
+image::images/data-model-3.png[All the reachable objects in your Git directory.]
 
 ==== Object Storage
 
--- a/book/10-git-internals/sections/objects.asc
+++ b/book/10-git-internals/sections/objects.asc
@@ -328,7 +328,7 @@ $ find .git/objects -type f
 
 If you follow all the internal pointers, you get an object graph something like this:
 
-.All the objects in your Git directory.
+.All the reachable objects in your Git directory.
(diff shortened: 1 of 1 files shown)


saadlu:
@Ben done it for the first time. Hope that's all that is required.

book/10-git-internals/sections/objects.asc
```diff
@@ -328,7 +328,7 @@ $ find .git/objects -type f
 
 If you follow all the internal pointers, you get an object graph something like this:
 
-.All the objects in your Git directory.
+.All the reachable objects in your Git directory.
 image::images/data-model-3.png[All the objects in your Git directory.]
```
ben: Can you update the square-bracket text as well? I forget which is which, but the text after the `.` and the text inside the `[`square brackets`]` are both important. One is the figure caption, the other is alt-text when rendered as HTML.

saadlu: my bad. Should be done now. Added as additional commit. Was wondering how to add the patch as an amend to the first one. Need to do some reading to know how (if it's possible).

ben: It's possible, but not necessary. There are some projects on GitHub that require every commit to be pristine, but this is not one of them. :)

PR #729, opened 2017-02-24
Update GitLab images

pedrorijo91:
From https://github.com/git/git-scm.com/pull/934


ben:
Ooh, nice! Since these aren't being printed, though, we'd like to avoid the user having to download huge binaries (both on the git-scm website and in the ebooks). Can you resize them to 800px wide and force-push?

pedrorijo91:
done @ben

ben:
✨ Thanks!

Issue #730, opened 2017-02-24
Suggest updating Apache configuration


daviesj:
I recommend updating the Apache configuration example on the [Smart HTTP](../blob/master/book/04-git-server/sections/smart-http.asc) section. Using the new authorization system in Apache 2.4 would simplify the configuration and remove the dependency on `mod_rewrite`. In fact, the `Order`, `Allow`, and `Deny` directives [are deprecated](https://httpd.apache.org/docs/2.4/howto/access.html). This change would not be backward compatible with Apache 2.2, but 2.4 has been out quite a while now and even Debian stable has been using Apache 2.4 for almost a couple years now.

ben:
Closed with #731.

PR #731, opened 2017-02-24
Update and simplify Apache config

daviesj:
Use `Require expr` feature of `mod_authz_core` to remove the need for mod_rewrite and simplify the entire configuration. 

This is my idea to fix issue #730. The single `Require expr ...` line replaces all of the rewrite rules and the Order/Deny/Satisfy lines.

Someone who knows more about Git than I do should probably verify that this configuration does what is needed. It seems to work for me, but I am much more knowledgeable about Apache configuration than Git's inner workings, so my testing of it may not have covered all cases.


ben:
Nice, thanks!

Issue #732, opened 2017-03-02
x86 and x64 Paths are switched


Qwampa2:
I think the x86 and x64 Paths are Switched at the following page:
https://git-scm.com/book/en/v2/Getting-Started-First-Time-Git-Setup
Take a quick look

YueLinHo:
Duplicate of issue #443

PR #707, opened 2017-01-16
Correct reversed x64/x86 commands

ericbarch:

PR #671, opened 2016-10-19
Update first-time-setup.asc

glasbrenner:
Exchange the x64 with the x86 system headline that desribes the usage of another text editor on a windows system.

Issue #443, opened 2015-11-03
Notepad++ x86 and x64 paths swapped


exbarboss:
First Time Git Setup chapter has error with path to Notepad++ on x86 and x64 versions of Windows.
It says:
On x86 system "'C:/Program Files/Notepad++...
On x64 system "'C:/Program Files(x86)/Notepad++...
Should be vise versa.

i-give-up:
No, the information there is correct. As far as I know, Notepad++ is only available as a 32-bit program, so it installs to

- `C:/Program Files` on x86 system (32-bit Windows)
- `C:/Program Files (x86)` on x64 system (64-bit Windows)

exbarboss:
You are right. I was under impression that I'm using x64 Notepad. Sorry for confusion.

PR #733, opened 2017-03-04
first-time-setup.asc. Editor commands. x86 V x64 swap.

JohnLukeBentley:
Swap commands for `git config --global core.editor ...` to properly
correspond to their respective introductions: "On a xYY system".

Mere errata.

### DIFF
--- a/book/01-introduction/sections/first-time-setup.asc
+++ b/book/01-introduction/sections/first-time-setup.asc
@@ -55,12 +55,12 @@ While on a Windows system, if you want to use a different text editor, such as N
 On a x86 system
 [source,console]
 ----
-$ git config --global core.editor "'C:/Program Files/Notepad++/notepad++.exe' -multiInst -nosession"
+$ git config --global core.editor "'C:/Program Files (x86)/Notepad++/notepad++.exe' -multiInst -nosession"
 ----
 On a x64 system
 [source,console]
 ----
-$ git config --global core.editor "'C:/Program Files (x86)/Notepad++/notepad++.exe' -multiInst -nosession"
+$ git config --global core.editor "'C:/Program Files/Notepad++/notepad++.exe' -multiInst -nosession"
 ----
 
 [NOTE]


lazarljubenovic:
See https://github.com/progit/progit2/issues/443.

JohnLukeBentley:
@lazarljubenovic thanks.

@i-give-up wrote in that thread:
> No, the information there is correct. As far as I know, Notepad++ is only available as a 32-bit program, so it installs to

>    C:/Program Files on x86 system (32-bit Windows)
>    C:/Program Files (x86) on x64 system (64-bit Windows)

Which would be correct if Notepad++ was only available as a 32-bit program, and the change suggested by my commit would be wrong.

However, the state of affairs today, and this is probably a recent thing, is that a 64-bit Notepad++ is available. It gets more complicated: the 64-bit version doesn't support most plugins. So in effect, for the immediate future, one should generally continue to only use the 32 bit version.

How about I push two more commits:
* Revert my prior commit; and
* Change the text to make it clear that the commands only refer to the 32 bit version. For example (my suggested change from the master version is emphasized):

> While on a Windows system, if you want to use a different text editor, such as
Notepad++, you can do the following *(when installing the 32 bit version of Notepad)*:
> On a x86 system 
>      $ git config --global core.editor "'C:/Program Files/Notepad++/notepad++.exe' -multiInst -no
> On a x64 system 
>       $ git config --global core.editor "'C:/Program Files (x86)/Notepad++/notepad++.exe' -multiIn

?

ben:
We have to consider the most common case for a reader who's on Windows. As of right now, it seems unlikely that anyone except plugin authors will be running 64-bit Notepad++, so the existing text is still the most relevant form.

Regardless of what happens with Notepad++, it's probable that we'll always live in a world with 32-bit editors running on 64-bit systems, so the general form of this text is still useful to the majority of readers. So once Notepad++ can recommend the 64-bit version, we might need to mention something else that ships as 32-bit to keep the usefulness of the text.

JohnLukeBentley:
Yes, I agree that the point of the text is less about Notepad++ specifically and more about providing example commands for, what is likely to be for the immediate future, 32 bit versions of editors.

However, in the current form the text lends itself to being misread. Evidently folk can, as I did, misread:

A. How you install 32 bit software on a 32 bit and 64 bit system (which is the intention); for
B. Different commands for 32 bit versions and 64 bit versions of software.

The evidence is in this sort of pull request being reopend, as at #671, #707, #732, arguably to the shame of all those pull requesters, including myself, given they didn't sufficiently search before hand. Shameful or not, the misreading seems to be recurrent.

So my suggestions remains: provide some kind of text to guard against that misreading. Something like "For installing 32 bit versions of ..." or some other guard.

This is also a matter I think you, @ben, can just make a call on one way or the other; and if you do decide to make changes to the text it's probably better coming from you.

ben:
@JohnLukeBentley, would you mind taking a look at #780?

PR #734, opened 2017-03-05
Modify command to change git user's login shell

jhhayashi:
Add additional information about chsh command, and simplify the command to change the login shell.


ben:
✨ Thanks!

PR #740, opened 2017-03-21
bzr-fastimport

aollier:
There was an error about the package to install with RHEL to import a Bazaar repository into a Git repository.
The good one is `bzr-fastimport`.


ben:
Brilliant. Thanks!

PR #741, opened 2017-03-25
show pictures in the text

Guanghui5:


jnavila:
Hello,

The management of picture does not work this way. The files in the `sections` folder are included in the files at the chapter level. These files are located at the `book/XX-title` level and from these locations the path to the images is correct, even if Github is unable to find them. The Github processing for asciidoc is only meant for standalone files, not for complex structures such as a book. If you want to read the book in its compile form please go to https://git-scm.com/book

Your change would break the building scripts that assemble the files.

PR #745, opened 2017-04-20
Improve comment about what the '**' glob pattern does.

chaitanyagurrapu:
Improve comment about what the '**' glob pattern does.

### DIFF
--- a/book/02-git-basics/sections/recording-changes.asc
+++ b/book/02-git-basics/sections/recording-changes.asc
@@ -246,7 +246,7 @@ build/
 # ignore doc/notes.txt, but not doc/server/arch.txt
 doc/*.txt
 
-# ignore all .pdf files in the doc/ directory
+# ignore all .pdf files in the doc/ directory or in the sub directories of doc/
 doc/**/*.pdf
 ----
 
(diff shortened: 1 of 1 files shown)


book/02-git-basics/sections/recording-changes.asc
```diff
@@ -246,7 +246,7 @@ build/
 # ignore doc/notes.txt, but not doc/server/arch.txt
 doc/*.txt
 
-# ignore all .pdf files in the doc/ directory
+# ignore all .pdf files in the doc/ directory or in the sub directories of doc/
```
ben: Hmm. What about this instead?

> `# ignore all .pdf files anywhere inside the doc/ directory`

chaitanyagurrapu: I think it is nicer to have the word "sub directory" in there somewhere. To make the association with '**' stronger and remove any ambiguity. After all that is what it is all about, which you might not know if you were seeing it for the first time.

ben: OK, how about this then:

> `# ignore all .pdf files in the doc/ directory or any of its subdirectories`

mdkalish: Maybe this is logically "more true"? :D

> \# ignore all .pdf files in the doc/ directory ~or~ and any of its subdirectories

chaitanyagurrapu:
Updated accordingly.

ben:
👍 Thanks!

PR #748, opened 2017-05-09
fix #744

### LINKED ISSUE #744: PDF does not have page numbers
Title says it all :)

delta4d:
add page numbers when generating pdf

there are 2 ways to add page numbers

1. add `:pagenums:` inside progit.asc
2. change Rakefile. `asciidoctor-pdf -a pagenums progit.asc`

or simply run `asciidoctor-pdf -a pagenums progit.asc 2>/dev/null` instead of  `rake book:build`

I chose the 1st, because it is more general. not sure which is better.

### DIFF
--- a/progit.asc
+++ b/progit.asc
@@ -4,6 +4,7 @@ Pro Git
 :docinfo:
 :toc:
 :toclevels: 2
+:pagenums:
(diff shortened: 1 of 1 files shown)


ben:
Nice, thanks!

PR #752, opened 2017-05-19
corrected the representation of the combination of keys "alt + →"

aollier:
Hello,

I realized that there are some problems with the PDF version of progit in the section GitLab due to the xml entity representing the right arrow.
The problems were :
* the link to bitnami was not correctly formed
* in the french version, the apostrophe was replaced by "&amp;#8217;"
* the right-arrow doesn't appear as it should be.

Here are the changes I made :

* replaced the '-' delimiter by a '+'
* replaced the xml entity that represents the right arrow by the symbol of the right arrow ('→')

Adrien Ollier


ben:
I suppose we already use some utf-8 characters in other chapters, so this shouldn't introduce any errors into the build toolchain. Thanks!

PR #753, opened 2017-05-20
Changes for chapter 7

sivaraam:
This PR adds a few changes to chapter 7 of the book that seemed good to me ;) Few more changes are to would possibly be made in a following PR. There's one change that I wanted to make in this PR but held it back to get comments regarding it. It's regarding the lines quoted below found in the _Changing the Last Commit_ subsection of the _Rewriting History_ section,

> You stage the changes you want by editing a file and running `git add` on it or `git rm` to a tracked file, and the subsequent git commit --amend takes your current staging area and makes it the snapshot for the new commit.

As far as I know, amending takes things in the current staging area and *adds* it to the snapshot of the last commit. So shouldn't the above sentence be rephrased as follows,
> You stage the changes you want by editing a file and running `git add` on it or `git rm` to a tracked file, and the subsequent git commit --amend takes your current staging area and *adds it to the last snapshot* making it the snapshot for the new commit.

### DIFF
--- a/book/07-git-tools/sections/searching.asc
+++ b/book/07-git-tools/sections/searching.asc
@@ -111,8 +111,9 @@ If you need to be more specific, you can provide a regular expression to search 
 ===== Line Log Search
 
 Another fairly advanced log search that is insanely useful is the line history search.
-This is a fairly recent addition and not very well known, but it can be really helpful.
-It is called with the `-L` option to `git log` and will show you the history of a function or line of code in your codebase.
+This can be really helpful.
+It is called with the `-L` option to `git log`.
+It will show you the history of a function or line of code in your codebase.
 
 For example, if we wanted to see every change made to the function `git_deflate_bound` in the `zlib.c` file, we could run `git log -L :git_deflate_bound:zlib.c`.
 This will try to figure out what the bounds of that function are and then look through the history and show us every change that was made to the function as a series of patches back to when the function was first created.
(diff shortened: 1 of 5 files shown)


book/07-git-tools/sections/searching.asc
```diff
@@ -111,8 +111,9 @@ If you need to be more specific, you can provide a regular expression to search
 ===== Line Log Search
 
 Another fairly advanced log search that is insanely useful is the line history search.
-This is a fairly recent addition and not very well known, but it can be really helpful.
-It is called with the `-L` option to `git log` and will show you the history of a function or line of code in your codebase.
+This can be really helpful.
+It is called with the `-L` option to `git log`.
+It will show you the history of a function or line of code in your codebase.
```
ben: The structure here is kind of stilted. Instead of these three sentences, what about just one:

> Simply run `git log` with the `-L` option, and it will show you the history of a function or line of code in your codebase.

sivaraam: Done :smile:

sivaraam:
@ben Any thoughts about the question in the description ?

ben:
Ah, right, I somehow missed that at first. Your change sounds like an improvement. Go ahead and include it here, and we'll get this merged!

sivaraam:
Made the change but with a little modification. BTW, sorry about that grammatical error in the commit message. Should have been careful.

Issue #747, opened 2017-04-24
PDF link missing?


Vangelis66:
> The easiest way is simply to let us do it. A robot is standing by to look for new work on the main branch 
> and automatically build it for everyone.
> 
> You can find the current builds on http://git-scm.com/book.

... Apologies for being ...thick (!), but once I visit referenced link I fail to spot "current builds"; 
I distinctly remember a pdf link being available there, last time I visited I fetched file "**progit-en.1067.pdf**" with a PDF modification timestamp of 20160309 (still on disk today). Is this link not available anymore? 
FWIW, https://github.com/progit/progit2/releases only provide **progit-2.0.0.pdf** with a PDF modification timestamp of 20141021. 

EDIT: Thanks to the Internet Archive, it appears the download links were removed after 20170309 (last capture with the links showing up; PDF link points to file **progit-en.1084.pdf** with a 20160323 PDF timestamp) ; so, if this was intentional, references on README.asc should be removed now...

jnavila:
True, true. Much of the work took place in the git-scm repo, but we forgot to update the README here.

Vangelis66:
> Much of the work took place in the git-scm repo

So I've searched and indeed found https://github.com/git/git-scm.com/pull/983 that was merged on 20170311; 
is that a permanent removal and if not, any speculation as to when the PDF link'll be restored? 
Have always been a fan of the PDF format of the Git Book ...
Thanks.

jnavila:
Unfortunately, the building robot is no longer available. We replaced it with a direct publishing to the git-scm website, but all ebooks are missing. I didn't find a simple way to generate the pdf. What you can do is build the PDF from the repo, the scripts for this are quite accessibles.

brunen9:
Here are ebooks builded at 2016/05/01: https://drive.google.com/drive/folders/0BzAMsjLtoLg-NFFHdGdBb3Qyd1E?usp=sharing

Vangelis66:
@brunen9 

Many thanks for those 👍 
Being on Windows (Vista), it's quite cumbersome to generate the PDF version locally 
from the repo scripts ;-( ; so your offerings are highly appreciated! 

Cheers

MichaelTiernan:
Thanks to brunen9 for providing the prebuilt books.
May I point out that the README file in the source doesn't specify the tools the building the books requires.

I've added asciidoctor to my system and have yet to find the "bundle" command.

It would be nice/important to have a prerequisite section.

[...]

PR #754, opened 2017-05-26
Update README.asc

cnguyen83:
Addressing issues raised in #747 . Clarifies that the git website only hosts an HTML version and that a local build can be generated with a different `asciidoctor` command option.

### DIFF
--- a/README.asc
+++ b/README.asc
@@ -17,9 +17,8 @@ See link:CONTRIBUTING.md[the Contributing document] for more information.
 There are two ways to generate e-book content from this source code.
 
 The easiest way is simply to let us do it.
-A robot is standing by to look for new work on the main branch and automatically build it for everyone.
 
-You can find the current builds on http://git-scm.com/book[].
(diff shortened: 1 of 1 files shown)


cnguyen83:
Question: is there another shortcut asciidoctor command to generate the epub version? That could also be added in, instead of just the PDF generation command.

aollier:
For EPub format:
```
bundle exec asciidoctor-epub3 progit.asc
```

For Mobi (kf8):
```
bundle exec asciidoctor-epub3 -a ebook-format=kf8 progit.asc
```

cnguyen83:
Thanks @aollier ! Since `bundle` was the issue in #747 I incorporated your commands without `bundle exec`, and it worked. I have local issues building, but they seem to be on the asciidoctor end, so not reflecting that in the README.

aollier:
Sorry I had read too fast, I didn't realize that `bundle` was an issue.
Glad nevertheless to have helped you.

ben:
👍🏼 thanks!

PR #760, opened 2017-05-28
feat(refs.asc): improve code example so that it works on emulated cmd…

ilkercat:
… lines on windows.

On emulated command line tool on Windows the quotation marks are saved into saved as well. This led to issues when trying to log the Git history.


ben:
Nice. Thanks!

PR #761, opened 2017-05-30
No package getopt on debian

Pherserk:
No package getopt on debian: https://packages.debian.org/search?suite=jessie&arch=i386&mode=filename&searchon=contents&keywords=getopt

### DIFF
--- a/book/01-introduction/sections/installing.asc
+++ b/book/01-introduction/sections/installing.asc
@@ -86,7 +86,7 @@ In order to be able to add the documentation in various formats (doc, html, info
 [source,console]
 ----
 $ sudo dnf install asciidoc xmlto docbook2X getopt
-$ sudo apt-get install asciidoc xmlto docbook2x getopt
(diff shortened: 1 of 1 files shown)


ben:
Can you confirm that this new command line will make it so you can build the Git documentation from source?

jnavila:
On my Debian Testing system, there's no `gnulib` package installed. `/usr/include/getopt.h` is coming from the package `libc6-dev`.

ben:
@Pherserk, we're coming up on 9 months since you submitted this. Any chance you can confirm that it works?

jnavila:
`gnulib` is a subset of libc for embedded systems, I'm pretty sure that this isn't needed. OTOH, `libc6-dev` is installed as a dependency of `build-essential` which should be listed.

ben:
Well, it looks like the source fork for this PR is gone. I'm going to close this out for now, let's see if anyone else runs up against this.

PR #768, opened 2017-06-16
used a real password for the Gmail account

aollier:
I propose this update to illustrate the use of a really strong password.


ben:
Yeah, okay. 👍🏼

PR #771, opened 2017-07-06
Update gitweb apache example for apache2.4

thcipriani:
In apache2.4 mixing `Options` with a `+` or a `-` with those without
will cause startup to abort. This patch updates the example gitweb
apache virtualserver to a valid configuration for apache2.4.


sajiblal:
2.4

sajiblal:
#771

PR #776, opened 2017-07-26
replaced "index" with "staging area" in section 2.2

td2014:
While reading this excellent book on Git,  in section 2.2 (Recording Changes...), found a slightly confusing use of the term "index" in the section on removal of files:

"The next time you commit, the file will be gone and no longer tracked. If you modified the file and added it to the index already ..."

I don't believe "index" was discussed in the prior sections, so  I replaced "index" with "staging area" which I think is what index is referring to.  This seems to read a little more consistently or clearly relative to the rest of the material.  I hope this is helpful.  Thanks.

### DIFF
--- a/book/02-git-basics/sections/recording-changes.asc
+++ b/book/02-git-basics/sections/recording-changes.asc
@@ -521,7 +521,7 @@ Changes to be committed:
 ----
 
 The next time you commit, the file will be gone and no longer tracked.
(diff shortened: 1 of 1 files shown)


ben:
Nice catch! Yeah, the terms "index" and "staging area" are sort of used interchangeably. Git itself uses "index" in the `git diff` output, but for prose we try to stay consistent with a less-confusing term. Thanks!

PR #777, opened 2017-08-04
Inverted x86 and x64

ThibaudLopez:
Corrected x86 and x64 that were inverted

### DIFF
--- a/book/01-introduction/sections/first-time-setup.asc
+++ b/book/01-introduction/sections/first-time-setup.asc
@@ -55,12 +55,12 @@ While on a Windows system, if you want to use a different text editor, such as N
 On a x86 system
 [source,console]
 ----
-$ git config --global core.editor "'C:/Program Files/Notepad++/notepad++.exe' -multiInst -nosession"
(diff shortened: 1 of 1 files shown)


ben:
😂 This is a common misconception, and further proof that

ben:
Microsoft _really_ made this confusing; take a look at [all the other people](https://github.com/progit/progit2/search?utf8=%E2%9C%93&q=program+files+%28x86%29&type=Issues) who have tried to correct this.

The underlying concept is that `C:\Program Files` contains applications that are native to the bitness of the Windows that's running; 32-bit on a 32-bit Windows, 64 on a 64. But to run a 32-bit app on 64-bit windows, you put the app in `C:\Program Files (x86)`, which doesn't exist on a 32-bit system. Notepad++ is a 32-bit app, so to install it on a 64-bit system, you put it in the "x86" folder.

So thanks for submitting this, but I have to close it, because the current version is confusingly correct. 😄

ThibaudLopez:
My bad, duh. I also failed to check similar pull requests previously closed. smh

PR #780, opened 2017-08-15
Resolve 32/64-bit confusion for Windows editors

ben:
This aims to address the confusion that leads to pull requests like #733, #732, #707, #671, and probably others. It aims to be more explicit with the`core.editor` instructions when it comes to 32-bit editors on a 64-bit system.

### DIFF
--- a/book/01-introduction/sections/first-time-setup.asc
+++ b/book/01-introduction/sections/first-time-setup.asc
@@ -53,7 +53,7 @@ $ git config --global core.editor emacs
 On a Windows system, if you want to use a different text editor, you must specify the full path to its executable file.
 This can be different depending on how your editor is packaged.
 
-In the case of Notepad++, a popular programming editor, the application is always shipped as a 32-bit executable.
+In the case of Notepad++, a popular programming editor, you are likely to want to use the 32-bit version, since at the time of writing the 64-bit version doesn't support all plug-ins.
 If you are on a 32-bit Windows system, or you have a 64-bit editor on a 64-bit system, you'll type something like this:
 
 [source,console]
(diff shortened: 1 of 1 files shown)


book/01-introduction/sections/first-time-setup.asc
```diff
@@ -50,14 +50,19 @@ If you want to use a different text editor, such as Emacs, you can do the follow
 $ git config --global core.editor emacs
 ----
 
-While on a Windows system, if you want to use a different text editor, such as Notepad++, you can do the following:
+On a Windows system, if you want to use a different text editor, you must specify the full path to its executable file.
+This can be different depending on how your editor is packaged.
+
+In the case of Notepad++, a popular programming editor, the application is always shipped as a 32-bit executable.
```
JohnLukeBentley: The following is not quite true ...
> In the case of Notepad++, a popular programming editor, the application is always shipped as a 32-bit executable.

As I wrote in #733  .....

> the state of affairs today, ...., is that a 64-bit Notepad++ is available. .... the 64-bit version doesn't support most plugins. (see https://notepad-plus-plus.org/download/v7.3.2.html)

> So in effect, for the immediate future, one should generally continue to only use the 32 bit version.

So I'd suggest something like the following wording: 
> This can be different depending on whether your editor or Windows version is 32 bit or 64 bit.

> In the case of Notepad++, a popular programming editor, you are likely to want to use the 32-bit version (as at the time of writing the 64-bit version doesn't support all the plug-ins ).

Otherwise this rest of the commit looks right.

ben: Nice, I'll use that. Thanks!

PR #783, opened 2017-08-24
Correct documentation for GIT_WORK_TREE based on git-config(1)

andrew-stripe:
The documentation for GIT_WORK_TREE incorrectly states that it will default to the parent directory of GIT_DIR. This contradicts the actual behavior and documentation of `core.worktree` in git-config(1): https://git-scm.com/docs/git-config#git-config-coreworktree

Do you think the same warning should be included in the documentation for `GIT_DIR`? I found this to be kind of a footgun when using `GIT_DIR`.


ben:
👣 🔫  Thanks!

PR #786, opened 2017-08-25
Correct section on checking out tags

### LINKED ISSUE #784: 2.6 Git Basics - Tagging — Inaccuracies
Hey!

I was reading up on tagging today and realized that the book states that you cannot check out tags directly. However, as detailed on the [githowto website](https://githowto.com/tagging_versions), running `git checkout v1.0` (or whatever your tag name) works just fine. I assume this is something that has changed since the section was written, and as such, it would make sense to update it accordingly, I think. If there is some information I'm missing here, however, please let me know.

Edit: Apologies for not reading the guidelines before submitting this. I could not, however, find this issue anywhere else. Furthermore, I was unable to find an actual pdf of this through a quick scan, but the source (in this repo) still has the same issue as reported above.

TheHeartmann:
This commit updates the section on checking out tags by replacing
the old text saying you can't really do it with a short paragraph
explaining how to do it.

---

Just a small tweak I found when reading up on tagging today. This removes the explanatory text after the command example too, which has not been replaced (partially because I wasn't sure exactly what to put there, and partially because it no longer seems as relevant, as it works the same way as when checking out commits), though I would be happy to do so if it is deemed necessary.

Fixes #784.

### DIFF
--- a/book/02-git-basics/sections/tagging.asc
+++ b/book/02-git-basics/sections/tagging.asc
@@ -211,13 +211,25 @@ Now, when someone else clones or pulls from your repository, they will get all y
 
 ==== Checking out Tags
 
-You can't really check out a tag in Git, since they can't be moved around.
-If you want to put a version of your repository in your working directory that looks like a specific tag, you can create a new branch at a specific tag with  `git checkout -b [branchname] [tagname]`:
+You can check out a tag through the `git checkout` command the same way you would with a commit:
 
 [source,console]
 ----
-$ git checkout -b version2 v2.0.0
-Switched to a new branch 'version2'
-----
+$ git checkout 2.0.0
+Note: checking out '2.0.0'.
 
-Of course if you do this and do a commit, your `version2` branch will be slightly different than your `v2.0.0` tag since it will move forward with your new changes, so do be careful.
+You are in 'detached HEAD' state. You can look around, make experimental
+changes and commit them, and you can discard any commits you make in this
+state without impacting any branches by performing another checkout.
+
+If you want to create a new branch to retain commits you create, you may
+do so (now or later) by using -b with the checkout command again. Example:
+
+  git checkout -b <new-branch-name>
+
+HEAD is now at 99ada87... Merge pull request #89 from schacon/appendix-final
+
+$ git checkout 2.0-beta-0.1
+Previous HEAD position was 99ada87... Merge pull request #89 from schacon/appendix-final
+HEAD is now at df3f601... add atlas.json and cover image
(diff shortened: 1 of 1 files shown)


book/02-git-basics/sections/tagging.asc
```diff
@@ -211,13 +211,25 @@ Now, when someone else clones or pulls from your repository, they will get all y
 
 ==== Checking out Tags
 
-You can't really check out a tag in Git, since they can't be moved around.
-If you want to put a version of your repository in your working directory that looks like a specific tag, you can create a new branch at a specific tag with  `git checkout -b [branchname] [tagname]`:
+You can check out a tag through the `git checkout` command the same way you would with a commit:
```
ben: So technically (and according to the output below) this puts you in "detached HEAD" state, which can be pretty confusing for complete newbies. We haven't even introduced the concept of HEAD, or the fact that branch pointers move when you make a commit. This might be too much information for the typical reader of this section.

Maybe something like this would be more appropriate:

> If you want to view the versions of files a tag is pointing to, you can do a `git checkout`, though this puts your repository into a state that isn't very useful:
>
> _[new example]_
>
> However, if you make changes and then create a commit, the tag will stay the same, and your commit won't be reachable from any branch. If you need to make changes, say if you're fixing a bug on an older version, you'll want to create a branch:
>
> _[old example]_

TheHeartmann: Ah, that's a very good point; I hadn't considered that at all. I agree with your suggestion, though I think the wording of it could benefit from a slight tweak to emphasise that the second paragraph talks about the detached HEAD state. Something more like:

>If you want to view the versions of files a tag is pointing to, you can do a `git checkout`, though this puts your repository in "detached HEAD" state, which has some ill side effects:
>
>*[new example]*
>
>In "detached HEAD" state, if you make changes and then create a commit, the tag will stay the same, but your new commit won't belong to any branch and will be unreachable, except for by the exact commit hash. Thus, if you need to make changes—say you're fixing a bug on an older version, for instance—you will generally want to create a branch:
>
>*[old example]*

(Also, I have to admit I'm a bit fuzzy on the exact effects of detached HEAD, so if anything above is wrong, please do not hesitate to correct me.)

ben: Nope, that sounds perfect. Make that change and we can get this merged!

TheHeartmann:
And done! I kept the original last line about being careful about making changes going forward on a new branch because I found it a helpful supplementation to the old example.

PR #788, opened 2017-08-28
New architecture of the repository for manual building of the book.

aollier:
The old architecture was needed for the building of the book by Atlas.
Atlas O'Reilly's building chain (https://atlas.oreilly.com) is broken.

- Images were moved into the new images root directory so that it is no more necessary to create it if it does not exist yet and copy all the images of the book into it.
- .gitignore, README.asc and Rakefile were adapted to integrate this change (simpler now)


ben:
Shame the output isn't as nice as Atlas, but this is definitely an improvement.

image

aollier:
The output was already like that before the move of the images. I don't know the reason. Maybe the images need to be processed ?

lazarljubenovic:
Since we're talking about images and the fact that they do not wok great atm, maybe it's the right time to include SVGs so anyone can edit them? They're currently in some paid Mac-only app if I recall correctly, and they do not seem that complex at all to re-create in SVG.

PR #802, opened 2017-09-16
There is no such problem in Git version 2.10.1

JunyiXie:
Git version 2.10.1 
```
Now suppose you did that in a branch. If you try to switch back to a branch where those files are still in the actual tree rather than a submodule – you get this error:

$ git checkout master
error: The following untracked working tree files would be overwritten by checkout:
  CryptoLibrary/Makefile
  CryptoLibrary/includes/crypto.h
  ...
Please move or remove them before you can switch branches.
Aborting
You can force it to switch with checkout -f, but be careful that you don’t have unsaved changes in there as they could be overwritten with that command.

$ git checkout -f master
warning: unable to rmdir CryptoLibrary: Directory not empty
Switched to branch 'master'
Then, when you switch back, you get an empty CryptoLibrary directory for some reason and git submodule update may not fix it either. You may need to go into your submodule directory and run a git checkout . to get all your files back. You could run this in a submodule foreach script to run it for multiple submodules.
```
 Did some tests and found that did not appear.

Are i wrong?

### DIFF
--- a/book/07-git-tools/sections/submodules.asc
+++ b/book/07-git-tools/sections/submodules.asc
@@ -874,33 +874,6 @@ Unpacking objects: 100% (11/11), done.
 Checking connectivity... done.
 ----
 
-Now suppose you did that in a branch.
-If you try to switch back to a branch where those files are still in the actual tree rather than a submodule – you get this error:
-
-[source,console]
-----
-$ git checkout master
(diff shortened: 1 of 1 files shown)


YueLinHo:
Wrong usage of PR. Please close it!

PR #808, opened 2017-09-20
Use square brackets for checkout [commit] [file]

lheckemann:
git checkout can be used to reset a file in the work tree to the current HEAD without specifying any further, so the commit parameter is optional.

### DIFF
--- a/book/07-git-tools/sections/reset.asc
+++ b/book/07-git-tools/sections/reset.asc
@@ -308,8 +308,8 @@ Pay especial attention to the 'WD Safe?' column – if it says *NO*, take a seco
 | `reset --soft [commit]` | REF | NO | NO | YES
 | `reset [commit]` | REF | YES | NO | YES
 | `reset --hard [commit]` | REF | YES | YES | *NO*
-| `checkout [commit]` | HEAD | YES | YES | YES
+| `checkout <commit>` | HEAD | YES | YES | YES
 | *File Level* | | | |
-| `reset (commit) [file]` | NO | YES | NO | YES
-| `checkout [commit] [file]` | NO | YES | YES | *NO*
+| `reset [commit] <file>` | NO | YES | NO | YES
+| `checkout [commit] <file>` | NO | YES | YES | *NO*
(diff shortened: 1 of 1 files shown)


book/07-git-tools/sections/reset.asc
```diff
@@ -311,5 +311,5 @@ Pay especial attention to the 'WD Safe?' column – if it says *NO*, take a seco
 | `checkout [commit]` | HEAD | YES | YES | YES
 | *File Level* | | | |
 | `reset (commit) [file]` | NO | YES | NO | YES
-| `checkout (commit) [file]` | NO | YES | YES | *NO*
+| `checkout [commit] [file]` | NO | YES | YES | *NO*
```
ben: 🤔 According to this convention, the `file` parameter isn't optional here. Maybe change this to `checkout [commit] (file)`?

rpjday: According to our recent discussions, that would be <file>, no? The man pages take it one step further, using the syntax [<branch>] for something optional, but that's probably getting a bit carried away.

ben: This part of the table is about file-specific operations using `reset` and `checkout`. If you don't include a file argument, then the behavior is specified elsewhere in the table.

rpjday: Sorry, didn't realize that commenting did strange things to square and angle brackets, which is why my note looks so weird. I was only suggesting that, based on recent conversations, a non-optional argument should be contained in angle brackets, correct?

ben: Oh right, yes, this would match up with your changes in other chapters.

@lheckemann, can you update your PR to surround optional arguments in `[square brackets]` and required-but-replaceable arguments in `<angle brackets>`?

lheckemann:
I think that's right now? This is just reminding me how awful the checkout/reset UI actually is…

ben:
Yeah, it's not ideal, but now this table is! Thanks!

PR #812, opened 2017-09-22
Change figure 101 caption

odises:
The caption of Figure 100 and Figure 101 is the same. I changed the 101 to "Link back to the new Pull Request in the closed Pull Request timeline" that I think is more related to the context.


ben:
Nice. Thanks!

PR #814, opened 2017-09-23
Variety of grammar/syntax/font changes and standardizations to Chapte…

rpjday:
…r 1.

Various cleanup tweaks including:

 * changing some commands/filenames to monospaced font
 * minor rewording and clarifications

### DIFF
--- a/book/01-introduction/1-introduction.asc
+++ b/book/01-introduction/1-introduction.asc
@@ -21,6 +21,6 @@ include::sections/help.asc[]
 
 === Summary
 
-You should have a basic understanding of what Git is and how it's different from the centralized version control system(s) you may have been using previously. 
+You should have a basic understanding of what Git is and how it's different from any centralized version control systems you may have been using previously.
 You should also now have a working version of Git on your system that's set up with your personal identity.
 It's now time to learn some Git basics.
--- a/book/01-introduction/1-introduction.asc
+++ b/book/01-introduction/1-introduction.asc
@@ -21,6 +21,6 @@ include::sections/help.asc[]
 
 === Summary
 
-You should have a basic understanding of what Git is and how it's different from the centralized version control system you may have previously been using.
+You should have a basic understanding of what Git is and how it's different from the centralized version control system(s) you may have been using previously. 
 You should also now have a working version of Git on your system that's set up with your personal identity.
 It's now time to learn some Git basics.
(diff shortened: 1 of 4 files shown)


book/01-introduction/1-introduction.asc
```diff
@@ -21,6 +21,6 @@ include::sections/help.asc[]
 
 === Summary
 
-You should have a basic understanding of what Git is and how it's different from the centralized version control system you may have previously been using.
+You should have a basic understanding of what Git is and how it's different from the centralized version control system(s) you may have been using previously.
```
ben: Looks like there's a trailing space here. Also, let's change this to say:

> how it's different from any centralized version control systems you may have been using previously.

book/01-introduction/sections/basics.asc
```diff
@@ -34,7 +34,8 @@ Most operations in Git only need local files and resources to operate – genera
 If you're used to a CVCS where most operations have that network latency overhead, this aspect of Git will make you think that the gods of speed have blessed Git with unworldly powers.
 Because you have the entire history of the project right there on your local disk, most operations seem almost instantaneous.
 
-For example, to browse the history of the project, Git doesn't need to go out to the server to get the history and display it for you – it simply reads it directly from your local database.
+For example, to browse the history of the project, Git doesn't need to go out to the server to get the history and display it for you – it simply reads it directly from your local database
+(commonly called the _object store_).
```
ben: First: the convention in this repo is to keep all sentences on a single line, and let your editor soft-wrap. That makes diffing easier – notice how it didn't highlight the deleted period on the first line here. 

Second: this is chapter 1. The number of technical terms we want to introduce here is as close to zero as possible, and the concept that a thing can have two names doesn't fit here.

Third: it's actually called the "object database."

Could you roll back this change?

book/01-introduction/1-introduction.asc
rpjday: Dumb question here ... just getting used to GitHub, can I do "git commit --amend" for this on my local system, then push it to my forked repo to the same branch, and redo the pull request? Is that how it works? I'm still reading the chapter on GitHub.

rpjday:
OK, I made the changes, pushed the additional commit. Is that sufficient?

ben:
That's _exactly_ what you should do. Thanks!

PR #815, opened 2017-09-23
Various grammar/font/clarification fixes to Chapter 2.

rpjday:
Tweaks include:

 - adjust some filenames/commands to use monospaced font
 - standardize on <> for replaceables, [] for optionals
 - Remove '$' in front of shell variables when only discussing them
 - Add "--oneline" option to log option table, given its popularity
 - Add NOTE clarifying "--all-match" option to git log
 - Minor rewording in a couple places for clarity

### DIFF
--- a/book/02-git-basics/sections/undoing.asc
+++ b/book/02-git-basics/sections/undoing.asc
@@ -7,8 +7,7 @@ Be careful, because you can't always undo some of these undos.
 This is one of the few areas in Git where you may lose some work if you do it wrong.
 
 One of the common undos takes place when you commit too early and possibly forget to add some files, or you mess up your commit message.
-If you want to redo that commit, make the additional changes you forgot, stage them, and commit
-again using the `--amend` option:
+If you want to redo that commit, make the additional changes you forgot, stage them, and commit again using the `--amend` option:
 
 [source,console]
 ----
--- a/book/02-git-basics/sections/undoing.asc
+++ b/book/02-git-basics/sections/undoing.asc
@@ -7,7 +7,8 @@ Be careful, because you can't always undo some of these undos.
 This is one of the few areas in Git where you may lose some work if you do it wrong.
 
 One of the common undos takes place when you commit too early and possibly forget to add some files, or you mess up your commit message.
-If you want to try that commit again, you can run commit with the `--amend` option:
+If you want to redo that commit, make the additional changes you forgot, stage them, and commit
+again using the `--amend` option:
 
 [source,console]
 ----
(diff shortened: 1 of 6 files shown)


book/02-git-basics/sections/undoing.asc
```diff
@@ -7,7 +7,8 @@ Be careful, because you can't always undo some of these undos.
 This is one of the few areas in Git where you may lose some work if you do it wrong.
 
 One of the common undos takes place when you commit too early and possibly forget to add some files, or you mess up your commit message.
-If you want to try that commit again, you can run commit with the `--amend` option:
+If you want to redo that commit, make the additional changes you forgot, stage them, and commit
+again using the `--amend` option:
```
ben: It may seem weird, but we've tried pretty hard to stick to the one-line-equals-one-sentence rule with the prose in this book. It makes the diffs more readable. Can you remove the line-break here?

book/02-git-basics/sections/viewing-history.asc
```diff
@@ -243,9 +243,17 @@ This command works with lots of formats – you can specify a specific date like
...
+[NOTE]
+====
+You can specify more than one instance of both the `--author` and `--grep` search criteria, which
+will limit the commit output to commits that match _any_ of the `--author` patterns and _any_
+of the `--grep` patterns; however, adding the `--all-match` option further limits the output to
+just those commits that match _all_ `--grep` patterns.
+====
+
+Another really helpful filter is the `-S` option (colloquially referred to as Git's ``pickaxe'' option)
+which takes a string and shows only those commits that changed the number of occurrences of that string.
```
ben: Same comment about line breaks.

Let's add a comma here:

> Another really helpful filter is the `-S` option (colloquially referred to as Git's ``pickaxe'' option), which takes a string argument, and shows only those commits that changed the number of occurrences of that string.

book/02-git-basics/sections/undoing.asc
rpjday: No, that's fine, I appreciate the value of that, will fix and resubmit.

rpjday:
OK, I've pushed a followup commit to revert the line breaking and add the comma. Do these additional commits show up as separate commits in the history, or is there a way to simply collapse them (a la "commit --amend) to look like an eventual single, correct commit and not make a mess of history?

ben:
Since nobody else has your branch checked out, you could do a `commit --amend` and then a `push --force` if you like, it's all the same to me. But just pushing up another commit doesn't bother me either, we're not too picky about the shape of the history in this repo.

PR #817, opened 2017-09-23
Various font/grammar/clarification tweaks to Chapter 4

rpjday:
Tweaks include:

 - Add hyphenation where appropriate
 - Standardize on <> for replaceables, [] for optionals
 - Add NOTE to mention deprecated FTP/FTPS protocols
 - Add NOTE to clarify "--local" clone option
 - Add NOTE to mention ssh-copy-id utility for SSH configuration

### DIFF
--- a/book/04-git-server/sections/protocols.asc
+++ b/book/04-git-server/sections/protocols.asc
@@ -3,15 +3,21 @@
 Git can use four major protocols to transfer data: Local, HTTP, Secure Shell (SSH) and Git.
 Here we'll discuss what they are and in what basic circumstances you would want (or not want) to use them.
 
+[NOTE]
+====
+In addition to the above, both FTP and FTPS protocols are supported, but those protocols
+are inefficient and deprecated, and won't be discussed here.
+====
+
 ==== Local Protocol
 
 (((protocols, local)))
-The most basic is the _Local protocol_, in which the remote repository is in another directory on disk.
+The most basic is the _Local protocol_, in which the remote repository is in another directory on the same disk.
 This is often used if everyone on your team has access to a shared filesystem such as an NFS mount, or in the less likely case that everyone logs in to the same computer.
 The latter wouldn't be ideal, because all your code repository instances would reside on the same computer, making a catastrophic loss much more likely.
 
 If you have a shared mounted filesystem, then you can clone, push to, and pull from a local file-based repository.
-To clone a repository like this or to add one as a remote to an existing project, use the path to the repository as the URL.
+To clone a repository like this, or to add one as a remote to an existing project, use the path to the repository as the URL.
 For example, to clone a local repository, you can run something like this:
 
 [source,console]
@@ -31,6 +37,12 @@ If you just specify the path, Git tries to use hardlinks or directly copy the fi
 If you specify `file://`, Git fires up the processes that it normally uses to transfer data over a network which is generally a lot less efficient method of transferring the data.
 The main reason to specify the `file://` prefix is if you want a clean copy of the repository with extraneous references or objects left out – generally after an import from another version-control system or something similar (see <<_git_internals>> for maintenance tasks).
 We'll use the normal path here because doing so is almost always faster.
+
+[NOTE]
+====
+The first syntax above implies using the `--local` option for the `git clone` command;
+see the man page for `git clone` to understand precisely what that means.
+====
 
 To add a local repository to an existing Git project, you can run something like this:
 
@@ -82,7 +94,7 @@ Instead of having to set up different URLs for these things, you can now use a s
 If you try to push and the repository requires authentication (which it normally should), the server can prompt for a username and password.
 The same goes for read access.
 
-In fact, for services like GitHub, the URL you use to view the repository online (for example, ``https://github.com/schacon/simplegit[]'') is the same URL you can use to clone and, if you have access, push over.
+In fact, for services like GitHub, the URL you use to view the repository online (for example, ``https://github.com/schacon/simplegit'') is the same URL you can use to clone and, if you have access, push over.
 
 ===== Dumb HTTP
 
@@ -145,23 +157,24 @@ Read <<_credential_caching>> to see how to set up secure HTTP password caching o
 (((protocols, SSH)))
 A common transport protocol for Git when self-hosting is over SSH.
 This is because SSH access to servers is already set up in most places – and if it isn't, it's easy to do.
-SSH is also an authenticated network protocol; and because it's ubiquitous, it's generally easy to set up and use.
+SSH is also an authenticated network protocol and, because it's ubiquitous, it's generally easy to set up and use.
 
-To clone a Git repository over SSH, you can specify ssh:// URL like this:
+To clone a Git repository over SSH, you can specify an `ssh://` URL like this:
 
 [source,console]
 ----
-$ git clone ssh://user@server/project.git
+$ git clone ssh://[user@]server/project.git
 ----
 
 Or you can use the shorter scp-like syntax for the SSH protocol:
 
 [source,console]
 ----
-$ git clone user@server:project.git
+$ git clone [user@]server:project.git
 ----
 
-You can also not specify a user, and Git assumes the user you're currently logged in as.
+In both cases above, if you don't specify the optional username,
+Git assumes the user you're currently logged in as.
 
 ===== The Pros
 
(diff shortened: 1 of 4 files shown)


book/04-git-server/sections/protocols.asc
```diff
@@ -3,15 +3,21 @@
...
+====
+In addition to the above, both FTP and FTPS protocols are supported, but those protocols
+are inefficient and deprecated, and won't be discussed here.
+====
+
 ==== Local Protocol
 
 (((protocols, local)))
-The most basic is the _Local protocol_, in which the remote repository is in another directory on disk.
+The most basic is the _Local protocol_, in which the remote repository is in another directory on the same disk.
```
ben: Well technically not, and the very next sentence will contradict this change. "Local clone" just means that Git is talking to the filesystem API to get the data from the remote repository, rather than network API. On my Mac I can do `cd /tmp; git clone /Volumes/mountedNetworkDisk/remote.git`. That'll use the "local" transport, but still technically result in network traffic.

I'm not sure the right way to convey all that in five words, but "on the same disk" doesn't quite do it. Maybe this:

> […] the remote repository is in another directory in the filesystem.

What do you think?

book/04-git-server/sections/protocols.asc
```diff
@@ -32,6 +38,12 @@ If you specify `file://`, Git fires up the processes that it normally uses to tr
 The main reason to specify the `file://` prefix is if you want a clean copy of the repository with extraneous references or objects left out – generally after an import from another version-control system or something similar (see <<_git_internals>> for maintenance tasks).
 We'll use the normal path here because doing so is almost always faster.
 
+[NOTE]
+====
+The first syntax above implies using the `--local` option for the `git clone` command;
+see the man page for `git clone` to understand precisely what that means.
+====
+
```
ben: I don't think this note is necessary. If the user has need of the `--local` flag, they're probably doing something else that would require a visit to the manpage, so they'll find it on their own.

It might be worth saying that the book isn't intended to be a slightly-more-readable version of the manpages. This is a prose description of how to use Git, which means we have to use some storytelling tools to teach. Making sure to mention every command-line flag that's related to what we're talking about doesn't accomplish the goal of making the user better at Git.

book/04-git-server/sections/protocols.asc
```diff
@@ -3,15 +3,21 @@
 Git can use four major protocols to transfer data: Local, HTTP, Secure Shell (SSH) and Git.
 Here we'll discuss what they are and in what basic circumstances you would want (or not want) to use them.
 
+[NOTE]
+====
+In addition to the above, both FTP and FTPS protocols are supported, but those protocols
+are inefficient and deprecated, and won't be discussed here.
+====
+
```
ben: I don't think we need this note here. By not talking about FTP, we effectively communicate that it's not a "major protocol," and that using it shouldn't be relied upon.

book/04-git-server/sections/protocols.asc
```diff
@@ -82,7 +94,7 @@ Instead of having to set up different URLs for these things, you can now use a s
 If you try to push and the repository requires authentication (which it normally should), the server can prompt for a username and password.
 The same goes for read access.
 
-In fact, for services like GitHub, the URL you use to view the repository online (for example, ``https://github.com/schacon/simplegit[]'') is the same URL you can use to clone and, if you have access, push over.
+In fact, for services like GitHub, the URL you use to view the repository online (for example, ``https://github.com/schacon/simplegit'') is the same URL you can use to clone and, if you have access, push over.
```
ben: 🤔 Why is it undesirable for this link to be clickable?

book/04-git-server/sections/protocols.asc
```diff
@@ -145,23 +157,24 @@ Read <<_credential_caching>> to see how to set up secure HTTP password caching o
...
 
 [source,console]
 ----
-$ git clone user@server:project.git
+$ git clone [user@]server:project.git
 ----
 
-You can also not specify a user, and Git assumes the user you're currently logged in as.
+In both cases above, if you don't specify the optional username,
+Git assumes the user you're currently logged in as.
```
ben: Line breaks, please.

book/04-git-server/sections/protocols.asc
```diff
@@ -185,7 +198,7 @@ This is a special daemon that comes packaged with Git; it listens on a dedicated
 In order for a repository to be served over the Git protocol, you must create the `git-daemon-export-ok` file – the daemon won't serve a repository without that file in it – but other than that there is no security.
 Either the Git repository is available for everyone to clone or it isn't.
 This means that there is generally no pushing over this protocol.
-You can enable push access; but given the lack of authentication, if you turn on push access, anyone on the internet who finds your project's URL could push to your project.
+You can enable push access but, given the lack of authentication, if you turn on push access, anyone on the internet who finds your project's URL could push to your project.
```
ben: Maybe:

> You can enable push access but, given the lack of authentication, anyone on the internet who finds your project's URL could push to your project.

book/04-git-server/sections/setting-up-server.asc
```diff
@@ -4,6 +4,13 @@
 Let's walk through setting up SSH access on the server side.
 In this example, you'll use the `authorized_keys` method for authenticating your users.
 We also assume you're running a standard Linux distribution like Ubuntu.
+
+[NOTE]
+====
+A good deal of what is described here can be automated by using the `ssh-copy-id` command,
+rather than manually copying and installing public keys.
+====
+
```
ben: Nice, I like this one.

book/04-git-server/sections/protocols.asc
rpjday: Fair enough, I just mentioned it since it's in the man page for "git clone" and readers might wonder why it's not mentioned in the book. Will remove.

book/04-git-server/sections/protocols.asc
rpjday: I'm not crazy about that wording, since it comes back to what I asked about before -- the difference between being in the same filesystem (allowing hardlinks), and being in a different filesystem but on the same disk. I'll just revert this back to the original wording for now, and ponder further.

book/04-git-server/sections/protocols.asc
rpjday: Point taken, will remove. On the other hand, I think I will go back at some point and add a note about the possibility of multiple .gitignore files, but simply say they're possible but outside the scope of the book so go read the man page. In a case like that, I think knowing there might be more than one is pretty important.

book/04-git-server/sections/protocols.asc
rpjday: Yup, better.

book/04-git-server/sections/protocols.asc
rpjday: Whoops, sorry, I had no idea that's what it meant, I thought it was a typo. Still learning asciidoc. will revert.

book/04-git-server/sections/setting-up-server.asc
rpjday: Removed line break here.

book/04-git-server/sections/protocols.asc
ben: Yeah, it's the square brackets that do it, and it still works inside quotes. Here's what the version on `master` looks like [rendered in HTML](https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols#_smart_http):

image

book/04-git-server/sections/protocols.asc
```diff
@@ -82,7 +82,7 @@ Instead of having to set up different URLs for these things, you can now use a s
 If you try to push and the repository requires authentication (which it normally should), the server can prompt for a username and password.
 The same goes for read access.
 
-In fact, for services like GitHub, the URL you use to view the repository online (for example, ``https://github.com/schacon/simplegit[]'') is the same URL you can use to clone and, if you have access, push over.
+In fact, for services like GitHub, the URL you use to view the repository online (for example, https://github.com/schacon/simplegit[]) is the same URL you can use to clone and, if you have access, push over.
```
ben: I still think the quotes are appropriate, and per my last comment, don't seem to cause rendering problems. Can you add them back in?

rpjday: Wait, are you saying that you tested that and, with the quotes, that is still rendered as a clickable link? Because that is most definitely not what is happening here, and it's not what you're doing throughout the rest of Chapter 4. If I put those quotes back in, that is now rendered simply as text, literally within smart double quotes and with the [] at the end showing up as part of the string; not a clickable link.
I can put those quotes back in at your request, but I can assure you, it will not be rendered (at least on my system) the way you think it is.

ben: Oh you're right. I was going off of the [HTML rendering](https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols#_smart_http), but the PDF rendering is indeed broken:

image

Sorry about that, we can merge it now.

PR #826, opened 2017-09-30
Minor grammar/clarification tweaks to Chapter 7, "Rewriting History"

rpjday:
Changes include:

 - Adding NOTE to emphasize rewriting history should be local.
 - Add recursive option to example about removing editor backup files.

### DIFF
--- a/book/07-git-tools/sections/rewriting-history.asc
+++ b/book/07-git-tools/sections/rewriting-history.asc
@@ -8,6 +8,13 @@ This can involve changing the order of the commits, changing messages or modifyi
 
 In this section, you’ll cover how to accomplish these very useful tasks so that you can make your commit history look the way you want before you share it with others.
 
+[NOTE]
+====
+The cardinal rule with Git is that, since so much work is local within your clone, you have a great deal of freedom to rewrite your history _locally_.
+However, once you push your work, it is a different story entirely, and you should consider pushed work as final.
+In short, you should never push your work until you're happy with it and ready to share it with the rest of the world.
+====
+
 [[_git_amend]]
 ==== Changing the Last Commit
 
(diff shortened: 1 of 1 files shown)


book/07-git-tools/sections/rewriting-history.asc
```diff
@@ -25,7 +32,7 @@ That drops you into your text editor, which has your last commit message in it,
 When you save and close the editor, the editor writes a new commit containing that message and makes it your new last commit.
 
 If you’ve committed and then you want to change the snapshot you committed by adding or changing files, possibly because you forgot to add a newly created file when you originally committed, the process works basically the same way.
-You stage the changes you want by editing a file and running `git add` on it or `git rm` to a tracked file, and the subsequent `git commit --amend` takes your current staging area and adds it to the last snapshot making it the snapshot of the last commit.
+Make the changes you think you forgot, stage those changes, and the subsequent `git commit --amend` _replaces_ the previous commit with your new, improved commit.
```
ben: Much better!

book/07-git-tools/sections/rewriting-history.asc
```diff
@@ -8,6 +8,13 @@ This can involve changing the order of the commits, changing messages or modifyi
 
 In this section, you’ll cover how to accomplish these very useful tasks so that you can make your commit history look the way you want before you share it with others.
 
+[NOTE]
+====
+The cardinal rule with Git is that, since so much work is local within your clone, you have a great deal of freedom to rewrite your history _locally_.
+However, once you push your work, it is a different story entirely, and you should consider pushed work as final.
+In short, you should never push your work until you're happy with it and ready to share it with the rest of the world.
+====
```
ben: I agree with your goal here, which is to call out more prominently that history-rewriting is risky. I hate to be so prescriptive, though. This is absolutely good advice _in most circumstances,_ but there are times when rewriting an already-pushed commit is the right way forward.

I kind of want to pair a note like this, which cautions about rewriting shared commits, with a new section on what to do when shared history has been rewritten in `origin`, and now your team all have repos in weird states. The note should come at the top, and probably the new section at the end.

Do you think you're up to that on this branch? Or should we back this note out and make a separate issue out of it?

book/07-git-tools/sections/rewriting-history.asc
```diff
@@ -284,7 +291,7 @@ Ref 'refs/heads/master' was rewritten
 
 The `--tree-filter` option runs the specified command after each checkout of the project and then recommits the results.
 In this case, you remove a file called `passwords.txt` from every snapshot, whether it exists or not.
-If you want to remove all accidentally committed editor backup files, you can run something like `git filter-branch --tree-filter 'rm -f *~' HEAD`.
+If you want to remove all accidentally committed editor backup files, you can run something like `git filter-branch --tree-filter 'rm -rf *~' HEAD`.
```
ben: The `-r` flag for `rm` just means "remove directories, too." It doesn't make sense in this context.

book/07-git-tools/sections/rewriting-history.asc
rpjday: I'll leave this note in here for now, but reword it to make it less draconian; then we can ponder how to expand on this more completely.

book/07-git-tools/sections/rewriting-history.asc
rpjday: Done.

PR #831, opened 2017-09-30
Rename "Binary Search" section to "Bisection" section.

rpjday:
While, yes, bisection uses binary search, it's more appropriate to
refer to it officially as bisection.


ben:
:thinking: Not totally sold on this, but I'm with you that "binary search" could be improved upon. What about "searching history" or "automated history search"?

rpjday:
Personally, I lean ever so slightly to "Bisection", but you can leave it as "Binary Search", it was never a big deal. You make the call.

ben:
Let's stick with "binary search," then. "Bisection" isn't part of the canon, and this would be the only prose Google result for that term.

PR #864, opened 2017-10-16
Remove erroneous double quote

rpjday:


ben:
Nice catch!

rpjday:
Not really, since I'm the one who put it there.

PR #916, opened 2017-11-04
'it' replaced by 'is' in 10th chap, objects section

jaspu:
'it' replaced by 'is' in chapter 10, section 'objects', line 8th, word 4th


ben:
Whoops! Nice catch!

PR #922, opened 2017-11-06
Remove "this can be confusing, because"

gmarmstrong:
The sentence "If you've staged all of your changes, `git diff` will give you no output." sounds nice by itself, and treats the behavior as a feature to be expected rather than a complication.


ben:
Excellent point. Thanks!

PR #924, opened 2017-11-07
systemd config file uses full path to git binary.

### LINKED ISSUE #923: Proposed systemd configuration is not valid
Using the file from [the book](https://git-scm.com/book/en/v2/Git-on-the-Server-Git-Daemon), I get:

```
● git-daemon.service - Start Git Daemon
   Loaded: error (Reason: Invalid argument)
   Active: inactive (dead)

Nov 07 13:53:48 name systemd[1]: [/etc/systemd/system/git-daemon.service:5] Executable path is not absolut
Nov 07 13:53:48 name systemd[1]: git-daemon.service: Service lacks both ExecStart= and ExecStop= setting. 
```

This is on Ubuntu 17.04 with systemd 232. (Yes, I'm aware of the Ubuntu packages `git-daemon-run` and `git-daemon-sysvinit`.)

Changing `git` to `/usr/bin/git` solves the problem; `ExecStop=` doesn't seem to be necessary after all.

reitzig:
Fixes #923

### DIFF
--- a/book/04-git-server/sections/git-daemon.asc
+++ b/book/04-git-server/sections/git-daemon.asc
@@ -30,7 +30,7 @@ Simply place a file in `/etc/systemd/system/git-daemon.service` with these conte
 Description=Start Git Daemon
 
 [Service]
-ExecStart=git daemon --reuseaddr --base-path=/srv/git/ /srv/git/
+ExecStart=/usr/bin/git daemon --reuseaddr --base-path=/srv/git/ /srv/git/
 
 Restart=always
 RestartSec=500ms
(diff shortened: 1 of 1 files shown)


book/04-git-server/sections/git-daemon.asc
```diff
@@ -50,6 +50,8 @@ You might have noticed that Git daemon is started here with `git` as both group
 
 Modify it to fit your needs and make sure provided user exists on the system.
 
+Also, make sure that the Git binary is indeed located at `/usr/bin/git` and change the path if necessary.
+
```
ben: Let's make this part of the paragraph just before it, I think it'll flow better that way.

reitzig: I wasn't sure about the formatting rules here. Since the two user-related sentences are split across separate paragraphs, it felt odd to attach something very different to one of the two.

ben: 🤔 

> You might have noticed that Git daemon is started here with git as both group and user.
> 
> Modify it to fit your needs and make sure provided user exists on the system.
> 
> Also, make sure that the Git binary is indeed located at /usr/bin/git and change the path if necessary.

You know what? Let's make all of these the same paragraph. Oh, and can you add the word "the" between "make sure" and "provided user"?

reitzig:
@ben Done. (Not sure if the commit itself triggers a notification; does it?)

ben:
✨

ben:
It does, but it's always nice to get a message telling me that you're ready for another review. Thanks, and congrats on your first PR!

PR #926, opened 2017-11-18
replace 'at the tip' by 'at the top'

ngandois:
I'm not fluent by I think there is a typo here


jnavila:
'tip' is correct:
http://www.wordreference.com/enfr/tip => bout, extrémité

PR #928, opened 2017-12-01
Update --recursive to --recurse-submodules

mdbrlh:
Update the --recursive option to --recurse-submodules for the clone command because --recursive seems to be depreciated in versions > 2.11.0


ben:
Nice catch. Thanks!

PR #938, opened 2017-12-18
MyLicense

anonymous6936:
This is my licence for my project . Facebook for Developers inc. You can ask me if you need to see what all you can do to help continue the flow of a Free Nation


anonymous6936:
Please continue building this idea. Together we can change the world.  Whether you believe it or not it's true. We must stay focused and not let our government control our issues and ways of fixing those issue for free. Stop leering them charge us to build our own lives up to what we want to be... True freedom is a beautiful thing".. :)

PR #942, opened 2017-12-23
fixed minor grammar mistake

TopView:
Guessing this should be from/to, as that's what a "trivial merge" would probably do:
so all of the files you _haven't_ changed in will be updated.   <=was
so all of the files you _haven't_ changed will be updated.       <=suggesting as simpler than the other option:
so all of the files you _haven't_ changed in it will be updated. <= possible alternative


ben:
Nice catch, thanks!

PR #957, opened 2018-01-08
indicate what running git from the terminal actually is for a newbie

tansaku:
Was working with a person new to git and they were confused about how to run git from the terminal as suggested in the installation process for mac


ben:
✨

PR #960, opened 2018-01-18
Add a missing step of objects manipulate procedure

multics:
If `refs/heads/master` was not updated with sha1 value of some commit, `git log` will complain. We need to include a step to update it.


multics:
sorry, my fault. This change is not necessary.

PR #968, opened 2018-02-05
Added Tagalog Links

severinolorillajr:
Filipino / Tagalog Translations


jnavila:
Thanks! I'll do the same for git-scm this evening (CET).

severinolorillajr:
Thanks. Will continue with the rest of the chapters then. Cheers!

PR #969, opened 2018-02-05
Rework files

jnavila:
This is rework of file hierarchy and names to adapt the book to the requirements of asciidoctor for ebooks and pdf, while retaining a working publishing toolchain at git-scm.com.

@deoren @rahrah I think I fixed the missing crosslinks from appendix C , which mysteriously haven't shown up with epubcheck. The ones in other chapters which passed through the transformation scripts were fixed as well.

I'd be very interested in a reliable automatic check of this stuff...


deoren:
> @jnavila: I'd be very interested in a reliable automatic check of this stuff...

I can't speak to how difficult it is to setup, but I've seen both Circle CI and Travis CI provide good feedback when changes were introduced to a source tree that broke builds. Perhaps that could be implemented for this repo as well.

Of course it might not help this particular problem unless there is a build flag that can be enabled to catch it.

jnavila:
@deoren We're already using travis-ci to compile the books so that they are available for download from git-scm.com . It works great. What I'm looking for is a utility to check the resulting document for failures.

I used epubcheck but it seems to miss some errors. That's better than nothing, but if something more reliable was available and if the pdf file could also be tested, that would make travis even more useful (specially for PRs).

deoren:
> @jnavila: @deoren We're already using travis-ci to compile the books so that they are available for download from git-scm.com . It works great. What I'm looking for is a utility to check the resulting document for failures.
> 
> I used epubcheck but it seems to miss some errors. That's better than nothing, but if something more reliable was available and if the pdf file could also be tested, that would make travis even more useful (specially for PRs).

Gotcha. Unfortunately I don't know of anything, but if I stumble across something I'll be sure to share.

deoren:
@jnavila Not sure if it's helpful, but these are the latest warnings generated emitted when generating the files:

```
ubuntu@ubuntu-1604-virtual-machine:~/Desktop/progit2$ rake
Converting to HTML...
asciidoctor: WARNING: book/07-git-tools/sections/rerere.asc: line 2: id assigned to section already in use: _rerere
asciidoctor: WARNING: book/10-git-internals/sections/refs.asc: line 68: id assigned to section already in use: _the_head
 -- HTML output at progit.html
Converting to EPub...
 -- Epub output at progit.epub
Converting to Mobi (kf8)...
 -- Mobi output at progit.mobi
Converting to PDF... (this one takes a while)
 -- PDF output at progit.pdf
```

jnavila:
@ben I'm going to merge this now, because otherwise, merges will rapidly become hairy.

rahrah:
Just checked with 80bd7447a57fe88b5.

All seems good.  Thank you!

The two double def of xrefs warnings have gone.

Can only see that unzipping epubs then parsing out xrefs and marrying up can be the only easy to go.  Doing this for html seems the only option, too.

Thanks for all this work!

ben:
Slightly late 👍 . This is awesome, thanks everybody!

PR #978, opened 2018-02-09
Update information about paid GitHub plans

brotherben:
The language here is outdated, as GitHub now offers unlimited private repositories with all paid plans. I've tried to keep the messaging as simple as possible – hopefully this is OK! 😅


jnavila:
Thanks. I learnt something today.

PR #984, opened 2018-02-11
Filipino translation of introduction.asc

Cyyy1998:


jnavila:
We don't accept translation is this project.

PR #993, opened 2018-02-14
Partial Translation in policy.asc from line 1 to to 240 - total of 1131 words

cordeta:


jnavila:
Wrong upstream repo...

PR #994, opened 2018-02-15
Prefaces

jnavila:
This PR adds the prefaces, dedications, licences and contributors to the ebooks and pdfs.


jnavila:
Nothing left to fix.

PR #998, opened 2018-02-19
Add go-git to Appendinx B: embedding

bzz:
Adds one more way of embedding Git in your application using  [go-git](https://github.com/src-d/go-git) project.

go-git is quite mature, well maintained and used in production by a multiple companies implementation of Git in Golang.

Please, let me know if you guys think it could be of any value for the readers.

PR #999, opened 2018-02-19
Add Dulwich to Appendinx B: embedding

bzz:
Adds one more way of embedding Git in your application using  https://www.dulwich.io project.

Dulwich is independent Python implementation of the Git file formats and protocols.

As it touches the same files, it's based on/includes a commit from #998 and so will be rebased on top of it. Only ee2d9dd needs to be reviewed in here.

### DIFF
--- a/book/B-embedding-git/sections/dulwich.asc
+++ b/book/B-embedding-git/sections/dulwich.asc
@@ -1,0 +1,44 @@
+=== Dulwich
+
+(((Dulwich)))((("Python")))
+There is also a pure-Python Git implementation - Dulwich.
+The project is hosted under https://www.dulwich.io/
+It aims to provide an interface to git repositories (both local and remote) that doesn't call out to git directly but instead uses pure Python.
+It has an optional C extensions though, that significantly improve the performance.
+
+Dulwich follows git design and separate two basic levels of API: plumbing and porcelain.
+
+Here is an example of using the lower level API to access the commit message of the last commit:
+
+[source, python]
+-----
+from dulwich.repo import Repo
+r = Repo('.')
+r.head()
+# '57fbe010446356833a6ad1600059d80b1e731e15'
+
+c = r[r.head()]
+c
+# <Commit 015fc1267258458901a94d228e39f0a378370466>
+
+c.message
+# 'Add note about encoding.\n'
+-----
+
+To print a commit log using high-level porcelain API, one can use:
+
+[source, python]
+-----
+from dulwich import porcelain
+porcelain.log('.', max_entries=1)
+
+#commit: 57fbe010446356833a6ad1600059d80b1e731e15
+#Author: Jelmer Vernooĳ <jelmer@jelmer.uk>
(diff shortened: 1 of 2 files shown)


ben:
Would you mind waiting until #998 merges and rebasing this? I'd like to review it on its own, and it's going to have a small merge conflict.

bzz:
Absolutely. Thank you for a kind guidance.

bzz:
@ben small doc on pure-python library was rebased, updated and is ready for another pass

book/B-embedding-git/sections/dulwich.asc
```diff
@@ -0,0 +1,44 @@
+=== Dulwich
+
+(((Dulwich)))((("Python")))
```
ben: Interesting. Does this index entry require the quotes?

book/B-embedding-git/sections/dulwich.asc
```diff
@@ -0,0 +1,44 @@
...
+It has an optional C extensions though, that significantly improve the performance.
+
+Dulwich follows git design and separate two basic levels of API: plumbing and porcelain.
+
+Here is an example of using the lower level API to access the commit message of the last commit:
+
+[source, python]
+-----
+from dulwich.repo import Repo
+r = Repo('.')
```
ben: Would you mind renaming this variable to `current_repo` or something similar? And maybe `c` should be called `commit`.

book/B-embedding-git/sections/dulwich.asc
bzz: Do not know, to be hones, I just copied it from libgit2 example. 
Digging deeper, https://github.com/progit/progit2/pull/471 seems to be a reason for those quotes.

Will be happy to remove those if you think there might be some side-effects, but I guess same logic should be applicable for [go-git](https://github.com/progit/progit2/blame/master/book/B-embedding-git/sections/go-git.asc#L3) as well.

Let me know if you want me to remove both in this PR

book/B-embedding-git/sections/dulwich.asc
bzz: Right now it's just 1-to-1 copy from https://www.dulwich.io/docs/#getting-started

Please, let me know if you still think it should be change and I will be happy to do so.

book/B-embedding-git/sections/dulwich.asc
ben: Alright, let's maybe clean them up in a separate PR then.

book/B-embedding-git/sections/dulwich.asc
ben: Hmm. Should we make this example more readable, or easier to update? Let's stick with Dulwich's example then.

book/B-embedding-git/sections/dulwich.asc
bzz: 👍 https://github.com/progit/progit2/pull/1096 created

PR #1006, opened 2018-02-25
Changed John to Jessica in Forked public project

snehalshekatkar:
$ git request-pull origin/master myfork should produce commits by Jessica and not by John. Fixed the first appearence of John.


lazarljubenovic:
I've always wnodered who came up with names that start on J. What happened to Alice and Bob?!

schacon:
@lazarljubenovic if you're really wondering, Jessica is my wife's name and John was her father's name. :)

PR #1011, opened 2018-03-04
git cat-file -p master^{tree} in other shells

Rafi993:
usage of `git cat-file -p master^{tree} command` in other environments as specified in this answer https://stackoverflow.com/a/17121954/2445295 thanks to @Mellbourn

### DIFF
--- a/book/10-git-internals/sections/objects.asc
+++ b/book/10-git-internals/sections/objects.asc
@@ -139,11 +139,12 @@ $ git cat-file -p master^{tree}
 The `master^{tree}` syntax specifies the tree object that is pointed to by the last commit on your `master` branch.
 Notice that the `lib` subdirectory isn't a blob but a pointer to another tree:
 
-**NOTE**: while using `master^{tree}` syntax if you have error try the following
-
-* On Windows, in PowerShell and cmd.exe, ^ is used for escaping. There, you can alternatively write git `cat-file -p master^^{tree}`
-* In zsh `^` is a globbing character so use them with quotes `git cat-file -p "master^{tree}"`
-* In bash the command works without quotes
+[NOTE]
+====
+* Depending on what shell you use, you may encounter errors when using the `master^{tree}` syntax
+* In PowerShell or CMD on Windows, the `^` character is used for escaping, so you have to double it to avoid this: `git cat-file -p mater^^{tree}`
+* If you're using ZSH, the `^` character is used for globbing, so you have to enclose the whole expression in quotes: `git cat-file -p "master^{tree}"`
+====
 
 [source,console]
 ----
--- a/book/10-git-internals/sections/objects.asc
+++ b/book/10-git-internals/sections/objects.asc
@@ -139,6 +139,12 @@ $ git cat-file -p master^{tree}
 The `master^{tree}` syntax specifies the tree object that is pointed to by the last commit on your `master` branch.
 Notice that the `lib` subdirectory isn't a blob but a pointer to another tree:
 
+**NOTE**: while using `master^{tree}` syntax if you have error try the following
+
+* On Windows, in PowerShell and cmd.exe, ^ is used for escaping. There, you can alternatively write git `cat-file -p master^^{tree}`
+* In zsh `^` is a globbing character so use them with quotes `git cat-file -p "master^{tree}"`
+* In bash the command works without quotes
+
 [source,console]
 ----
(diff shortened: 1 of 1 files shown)


book/10-git-internals/sections/objects.asc
```diff
@@ -139,6 +139,12 @@ $ git cat-file -p master^{tree}
 The `master^{tree}` syntax specifies the tree object that is pointed to by the last commit on your `master` branch.
 Notice that the `lib` subdirectory isn't a blob but a pointer to another tree:
 
+**NOTE**: while using `master^{tree}` syntax if you have error try the following
```
ben: Instead of just setting this in bold-face, let's use the Asciidoc "note" feature:

```asciidoc
[NOTE]
====
(Note content goes here)
====
```

This will render much nicer.

In addition, let's change the prose here to read like this:

```asciidoc
Depending on what shell you use, you may encounter errors when using the `master^{tree}` syntax.
In PowerShell or CMD on Windows, the `^` character is used for escaping, so you have to double it to avoid this: `git cat-file -p master^^{tree}`.
If you're using ZSH, the `^` character is used for globbing, so you have to enclose the whole expression in quotes: `git cat-file -p "master^{tree}"`.
```

Or something like that. It's a bit more idiomatically English.

book/10-git-internals/sections/objects.asc
```diff
@@ -139,6 +139,13 @@ $ git cat-file -p master^{tree}
 The `master^{tree}` syntax specifies the tree object that is pointed to by the last commit on your `master` branch.
 Notice that the `lib` subdirectory isn't a blob but a pointer to another tree:
 
+[NOTE]
+====
+* Depending on what shell you use, you may encounter errors when using the `master^{tree}` syntax
+* In PowerShell or CMD on Windows, the `^` character is used for escaping, so you have to double it to avoid this: `git cat-file -p mater^^{tree}`
+* If you're using ZSH, the `^` character is used for globbing, so you have to enclose the whole expression in quotes: `git cat-file -p "master^{tree}"`
```
ben: Interesting. Why did you set this using bullets?

image

Let's just write it as a single paragraph, I think that flows better.

Rafi993: sure

book/10-git-internals/sections/objects.asc
```diff
@@ -139,6 +139,11 @@ $ git cat-file -p master^{tree}
 The `master^{tree}` syntax specifies the tree object that is pointed to by the last commit on your `master` branch.
 Notice that the `lib` subdirectory isn't a blob but a pointer to another tree:
 
+[NOTE]
+====
+Depending on what shell you use, you may encounter errors when using the `master^{tree}` syntax. In PowerShell or CMD on Windows, the `^` character is used for escaping, so you have to double it to avoid this: `git cat-file -p mater^^{tree}`. If you're using ZSH, the `^` character is used for globbing, so you have to enclose the whole expression in quotes: `git cat-file -p "master^{tree}"`
```
ben: Sorry, one more request: could you reformat this with one sentence on each source-code line? It makes reviewing PRs _much_ easier in the future.

Rafi993: sure no problem

book/10-git-internals/sections/objects.asc
```diff
@@ -139,6 +139,15 @@ $ git cat-file -p master^{tree}
...
 
+[NOTE]
+====
+Depending on what shell you use, you may encounter errors when using the `master^{tree}` syntax.
+
+In PowerShell or CMD on Windows, the `^` character is used for escaping, so you have to double it to avoid this: `git cat-file -p mater^^{tree}`.
+
+If you're using ZSH, the `^` character is used for globbing, so you have to enclose the whole expression in quotes: `git cat-file -p "master^{tree}"`
+====
+
```
ben: Oh man, I hate to do this to you again, but I have one last request (I promise). Looking at how this renders in context:

image

The "here's what you should do, and here's what it looks like when you do it" flow is interrupted by this note block. Could you move it just below the code block?

Rafi993: Sure no problem

Rafi993: Hope these formatting rules are defined somewhere so that people could just read and follow it.

PR #1035, opened 2018-04-17
Number consistency

eyherabh:
Replaced 'it' with 'them' because it refers to 'workflows' which is in plural.


ben:
Yes, definitely better. Thanks!

PR #1044, opened 2018-04-29
Update first-time-setup.asc

erikgruener:
When i call git config --list i will receive the local config which does not have my user name saved yet, therefore it should be git config --global --list in line 86

### DIFF
--- a/book/01-introduction/sections/first-time-setup.asc
+++ b/book/01-introduction/sections/first-time-setup.asc
@@ -83,7 +83,7 @@ An example on a Windows system may include a prematurely terminated Git operatio
 
 ==== Checking Your Settings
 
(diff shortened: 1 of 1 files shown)


ben:
That's odd. When I'm in this repo, `git config --list --local` comes back with 15 lines, but the output from `git config --list` is 73 lines long. The [docs](https://git-scm.com/docs/git-config#_description) say:

> When reading, the values are read from the system, global and repository local configuration files by default, and options `--system`, `--global`, `--local` and `--file <filename>` can be used to tell the command to read from only that location (see FILES).

Do you have a `GIT_CONFIG` environment variable set, or maybe you're using an alias?

erikgruener:
I was just editing the wording to fit the text and no i do not have environmental variable set. Oh well cheers

PR #1063, opened 2018-05-30
Fix comma splices

vic-ma:
In two instances, a comma was used when other punctuation would have
been more appropriate to seperate the two independant caluses.


ben:
Nice. Thanks!

PR #1067, opened 2018-06-04
Get Version from git describe not travis envvar

rahrah:
Users should get version number from git rather than a
possibly non-existent environment variable.


rahrah:
Just a thought re having version numbers in user built versions regardless of having travis.

===R

jnavila:
Why not mix both sources? Get the string version from Travis, and if it fails, use git describe.

rahrah:
How does this look??

:-)

===Rich

PR #1069, opened 2018-06-05
Fix issue 627: drop incorrect knock on older VCS

dtikhonov:


ben:
Yeah, the world has moved on since that paragraph was relevant. Plus, we're in chapter 3, so we don't need to sell people on Git at this point. Thanks!

PR #1072, opened 2018-06-06
Make some stylistic tweaks to Chapter 1, Introduction

rpjday:
Stylistic/aesthetic adjustments, including:

 - use "--" for em-style dashes, not special characters
 - remove superfluous Oxford comma
 - add some commas where they're warranted
 - expand on the "git config" options

### DIFF
--- a/book/01-introduction/sections/about-version-control.asc
+++ b/book/01-introduction/sections/about-version-control.asc
@@ -30,7 +30,7 @@ RCS works by keeping patch sets (that is, the differences between files) in a sp
 (((version control,centralized)))
 The next major issue that people encounter is that they need to collaborate with developers on other systems.
 To deal with this problem, Centralized Version Control Systems (CVCSs) were developed.
-These systems, such as CVS, Subversion, and Perforce, have a single server that contains all the versioned files, and a number of clients that check out files from that central place. (((CVS)))(((Subversion)))(((Perforce)))
+These systems, such as CVS, Subversion and Perforce, have a single server that contains all the versioned files, and a number of clients that check out files from that central place. (((CVS)))(((Subversion)))(((Perforce)))
 For many years, this has been the standard for version control.
 
 .Centralized version control.
--- a/book/01-introduction/sections/about-version-control.asc
+++ b/book/01-introduction/sections/about-version-control.asc
@@ -30,7 +30,7 @@ RCS works by keeping patch sets (that is, the differences between files) in a sp
 (((version control,centralized)))
 The next major issue that people encounter is that they need to collaborate with developers on other systems.
 To deal with this problem, Centralized Version Control Systems (CVCSs) were developed.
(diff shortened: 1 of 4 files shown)


lazarljubenovic:
> remove superfluous Oxford comma

Do we have a standard for this in the book? This is a change which could be ping-ponged between two groups of users until the end of time.

rpjday:
I typically wouldn't submit a pull request for just that, but since there was other stuff being changed, I snuck it in. The standard I've seen (and used) for a while is that the Oxford comma is not used unless it actually clarifies the expression, which would be ambiguous otherwise. I like [this explanation](https://en.oxforddictionaries.com/explore/what-is-the-oxford-comma). But if Ben wants to toss it, he has my blessing.

nobozo:
On 6/6/2018 9:12 AM, Robert P. J. Day wrote:
> I typically wouldn't submit a pull request for just that, but since 
> there was other stuff being changed, I snuck it in. The standard I've 
> seen (and used) for a while is that the Oxford comma is not used unless 
> it actually clarifies the expression, which would be ambiguous 
> otherwise. I like this explanation 
> <https://en.oxforddictionaries.com/explore/what-is-the-oxford-comma>. 
> But if Ben wants to toss it, he has my blessing.

I vote for consistency. Either the Oxford comment should always
be used, or it should never be used. If there is a case where
it's necessary for clarification then this is a good argument
for always using it.

Jon

book/01-introduction/sections/about-version-control.asc
```diff
@@ -30,7 +30,7 @@ RCS works by keeping patch sets (that is, the differences between files) in a sp
 (((version control,centralized)))
 The next major issue that people encounter is that they need to collaborate with developers on other systems.
 To deal with this problem, Centralized Version Control Systems (CVCSs) were developed.
-These systems, such as CVS, Subversion, and Perforce, have a single server that contains all the versioned files, and a number of clients that check out files from that central place. (((CVS)))(((Subversion)))(((Perforce)))
+These systems, such as CVS, Subversion and Perforce, have a single server that contains all the versioned files, and a number of clients that check out files from that central place. (((CVS)))(((Subversion)))(((Perforce)))
```
ben: In this case, it seems like the liberal sprinkling of commas is doing us all a disservice. What if we repunctuate to something like this:

> These systems (such as CVS, Subversion, and Perforce) have a single server that contains all the versioned files, and a number of clients that check out files from that central place.

That way it gets more readable AND I get to keep my Oxford comma. 😁

rpjday: I'm fine with whatever you decide. Tweak the pull request, I'm in the middle of more wide-sweeping changes. :-)

ben: Sadly I cannot push to your fork. :(

rpjday: OK, I'll resubmit with that change. (I thought I checked the box that lets you edit my pull requests, is that not what that means?)

ben: Ha, I was making that _way_ harder than I needed to. 😊

PR #1078, opened 2018-06-09
signing.asc: Standardize on using "$" to represent command markup

rpjday:
I propose that *all* command markup (even single-line markup)
use a dollar sign to signify a command, just for consistency.
Otherwise, it looks kind of weird as it bounces back and forth
between the two.

### DIFF
--- a/book/07-git-tools/sections/signing.asc
+++ b/book/07-git-tools/sections/signing.asc
@@ -22,14 +22,14 @@ If you don't have a key installed, you can generate one with `gpg --gen-key`.
 
 [source,console]
 ----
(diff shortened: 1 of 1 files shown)


ben:
I'm with you for block-level examples, but inline it doesn't make a ton of sense. Compare these:

> Simply run the `foo bar baz` command, and you'll see that…

> Simply run the `$ foo bar baz` command, and you'll see that…

The monospace segment is meant to be a noun. Adding the shell prompt encumbers it with all the shell/console/bash-vs-zsh-vs-fish context, none of which is what we're trying to talk about.

PR #1081, opened 2018-06-11
Add hash-object to validate ruby hash function

### LINKED ISSUE #1080: hash-object is different from ruby's implemention
In https://github.com/progit/progit2/blob/master/book/10-git-internals/sections/objects.asc#object-storage, we use ruby to hash `what is up, doc?` and we get:
```
bd9dbf5aae1a3862dd1526723246b20206e5fc37
```

However, I hash this string with `hash-object`, from which I got a different hash:

```
$ echo "what is up, doc?" | git hash-object --stdin -w
7108f7ecb345ee9d0084193f147cdad4d2998293
$ git cat-file -p 7108f7ecb345ee9d0084193f147cdad4d2998293
what is up, doc?
```

Test on:
macOS 10.13.5
git version 2.15.1 (Apple Git-101)
config: 
```
[core]
	repositoryformatversion = 0
	filemode = true
	bare = false
	logallrefupdates = true
	ignorecase = true
	precomposeunicode = true
```

jingsam:
close #1080

### DIFF
--- a/book/10-git-internals/sections/objects.asc
+++ b/book/10-git-internals/sections/objects.asc
@@ -381,12 +381,13 @@ You can calculate the SHA-1 value of a string in Ruby by including the SHA1 dige
 => "bd9dbf5aae1a3862dd1526723246b20206e5fc37"
 ----
 
-Let's validate the hash with `git hash-object`. It shows that they are identical. Note that `echo` will add a trailing newline character in string, which is not we want in this case. `echo -n` can prevent from adding a trailing newline character.
+Let's compare that to the output of `git hash-object`.
+Here we use `echo -n` to prevent adding a newline to the input.
 
 [source,console]
 ----
->> $ echo -n "what is up, doc?" | git hash-object --stdin
-=> bd9dbf5aae1a3862dd1526723246b20206e5fc37
+$ echo -n "what is up, doc?" | git hash-object --stdin
+bd9dbf5aae1a3862dd1526723246b20206e5fc37
 ----
 
 Git compresses the new content with zlib, which you can do in Ruby with the zlib library.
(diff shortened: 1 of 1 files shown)


book/10-git-internals/sections/objects.asc
```diff
@@ -381,6 +381,14 @@ You can calculate the SHA-1 value of a string in Ruby by including the SHA1 dige
 => "bd9dbf5aae1a3862dd1526723246b20206e5fc37"
 ----
 
+Let's validate the hash with `git hash-object`. It shows that they are identical. Note that `echo` will add a trailing newline character in string, which is not we want in this case. `echo -n` can prevent from adding a trailing newline character.
```
ben: Could you reformat this so that each sentence is on its own line? It makes future PRs much easier to review.

Also, let's make it a bit more succinct. Something like this maybe?

```
Let's compare that to the output of `git hash-object`.
Here we use `echo -n` to prevent adding a newline to the input.
```

What do you think?

book/10-git-internals/sections/objects.asc
```diff
@@ -381,6 +381,14 @@ You can calculate the SHA-1 value of a string in Ruby by including the SHA1 dige
...
+
+[source,console]
+----
+>> $ echo -n "what is up, doc?" | git hash-object --stdin
+=> bd9dbf5aae1a3862dd1526723246b20206e5fc37
+----
```
ben: Here you've got a shell sample marked up like it was typed into `irb`, which won't work. Could you remove the `>>` and `=>` prefixes?

jingsam:
ping @ben

ben:
✨

PR #1097, opened 2018-07-27
HEAD~3 is equivalent to HEAD~~~, not HEAD^^^

hydnum:
Fixing a typo where you wrote HEAD^^^ instead of HEAD~~~ (occurs in two places).


ben:
Oh hey, nice catch. Thanks!

PR #1099, opened 2018-08-04
Update ssh-keygen with secure password encryption

peterwwillis:
I know the documentation here suggests you don't need a password, but I am suggesting this change for two reasons:

1) If the user does want a password, they will want to use the -o option. Not using the -o option makes the password practically worthless. See here for details: https://latacora.singles/2018/08/03/the-default-openssh.html

2) This documentation comes up on the top 10 Google search results for "how do I make an ssh key". It's likely that ssh users (even experienced ones) may use this guide to create new keys. It is important to get users into the habit of using the -o option, since it should be the default for the command anyway, but for whatever reason was never updated as such.

### DIFF
--- a/book/04-git-server/sections/generating-ssh-key.asc
+++ b/book/04-git-server/sections/generating-ssh-key.asc
@@ -36,8 +36,8 @@ d0:82:24:8e:d7:f1:bb:9b:33:53:96:93:49:da:9b:e3 schacon@mylaptop.local
 ----
 
 First it confirms where you want to save the key (`.ssh/id_rsa`), and then it asks twice for a passphrase, which you can leave empty if you don't want to type a password when you use the key.
-
-However, if you do use a password, make sure to add the `-o` option. You can also use the `ssh-agent` tool to prevent having to enter the password each time.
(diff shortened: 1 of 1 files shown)


peterwwillis:
Hi Ben, does this work?

book/04-git-server/sections/generating-ssh-key.asc
```diff
@@ -37,6 +37,8 @@ d0:82:24:8e:d7:f1:bb:9b:33:53:96:93:49:da:9b:e3 schacon@mylaptop.local
 
 First it confirms where you want to save the key (`.ssh/id_rsa`), and then it asks twice for a passphrase, which you can leave empty if you don't want to type a password when you use the key.
 
+However, if you do use a password, make sure to add the `-o` option. You can also use the `ssh-agent` tool to prevent having to enter the password each time.
+
```
ben: Can you split up these sentences onto their own lines? Also, I don't think we need a paragraph break here, so go ahead and delete line 39.

ben:
✨ Thanks!

PR #1101, opened 2018-08-09
Explain -o option in ssh-keygen

rahrah:
Add brief explanation of

ssh-keygen -o

option.


rahrah:
Hi,

Just felt that the `ssh-keygen -o` option needed a little explanation (especially as it seems to be of general importance to ssh users).

===Rich

ben:
Lovely. Thanks! 🦄

PR #1104, opened 2018-09-05
Clarify explanatory text for .gitignore

evanderiel:


jnavila:
to be precise, it's "ignore all file in any sub-directory named build".

ben:
I don't think any reader would see that and think "if I clone this to a directory called `build`, will the repo ignore itself?" It's a good change, and I'm going to merge it. Thanks!

PR #1116, opened 2018-10-24
change the describe of when should force the removal with the -f option

Tvirus:
whether the file is modified or has been added to the staging area, the -f option is needed to force the removal


ben:
Better, yes. Thanks!

PR #1123, opened 2018-11-12
removed unused images

aollier:
Those two images are not used either:
```bash
> git grep --fixed-strings images/emoji.png -- \*.asc # no result
> git grep --fixed-strings images/notifications.png -- \*.asc # no result
```


ben:
Nice.

PR #1127, opened 2018-11-18
TortoiseGit bug tracker

### LINKED ISSUE #1126: TortoiseGit bugtraq
Related to #541:

The current configuration of TortoiseGit (.tgitconfig) does not create links when the commit message contains something like `closes #XXX` or `fixes #XXX` where XXX is a bug ID.
[Closing issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/) is the way users should mention issues so we should use a pattern that recognizes bugs in this format.

Examples are:
`closes #541`, `fixes: #541`, `resolved #783`.

I know the pattern to use so if you agree with that I may open a Pull Request that fixes this.

aollier:
fixes #1126


ben:
:sparkles:

YueLinHo:
:smile:

PR #1143, opened 2018-12-06
interactive staging: update prompts to include "[]" for shortcuts

rpjday:
The newer form of interactive staging uses square brackets
to emphasize single-letter shortcuts.


ben:
Nice!

PR #1153, opened 2019-01-03
Clarify when is not recommended to rebase

### LINKED ISSUE #1152: Rebase if not published does not fit in continous integration mind
I am opening this issue to discuss the recommendation about not rebasing a branch if it has been published. There are several situations that the preferred workflow brings you to do it.

The general advise about rebasing is not to do it once you have pushed the changes to the remote making the changes public, but a clear distinction should be done. Is the branch intended to be public or you should be the only one working on that branch? Or is it published but in a private repository?

Let me write an example:

You have a big project, with a lot of tests for different platforms. To make sure that your master branch is working you have a CI system running on all the branches and on the pull requests. Because of the different platforms, the only way that has the developer to know that the work will not break master is to let the CI run the test. And the only way to do that is pushing the changes to the branch.

To me there are two possible workflows to work but both require rebase after pushing the changes:

- The developer creates a fork of the main repository and creates a branch to implement a new feature. When she wants to test the changes, she can create a pull request to the main repository and the CI runs the tests.
- The developer creates a branch in the main repository with a convention clearly identifying what developer is (developer_id_feature_name) to implement a new feature. When she wants to test the changes, she can create a pull request to the main repository and the CI runs the tests.

In both scenarios the only way for the CI system to see the changes is that the developer pushes the changes.
After the pull request, the developer sees that some test is failing, she forgot some code in one of the commits. What should she do? Rebase creating another branch because it was already published or rebase her branch?

To me when there is a Continous Integration system in place, the rule does not apply anymore, or at least not in the current form. The idea would be:
You can rebase if you have not pushed the changes or if it has been pushed but you are rebasing a private branch. Where private branch means that by convention no one should checkout or create commits on top of it.

Because CI systems are getting more and more important and most of the times will be the ones running the tests, I think that it is worth it to rephrase the recommendation in the book.

limdor:
This pull requests wants to fix #1152 regarding when a user should and should not be doing rebase.
I have form it in a way trying to minimize the additions in the book.
Another option would be to add some section about how git and rebase can play a role in continous integration systems.

I am not native english, it would be great if some native english could review it.

### DIFF
--- a/book/03-git-branching/sections/rebasing.asc
+++ b/book/03-git-branching/sections/rebasing.asc
@@ -137,7 +137,7 @@ image::images/interesting-rebase-5.png[Final commit history.]
 (((rebasing, perils of)))
 Ahh, but the bliss of rebasing isn't without its drawbacks, which can be summed up in a single line:
 
-*Do not rebase commits that exist outside your repository.*
+*Do not rebase commits that exist outside your repository and people may have based work on them.*
 
 If you follow that guideline, you'll be fine.
 If you don't, people will hate you, and you'll be scorned by friends and family.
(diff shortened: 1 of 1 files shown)


book/03-git-branching/sections/rebasing.asc
```diff
@@ -208,7 +208,7 @@ Or you could do it manually with a `git fetch` followed by a `git rebase teamone
 
 If you are using `git pull` and want to make `--rebase` the default, you can set the `pull.rebase` config value with something like `git config --global pull.rebase true`.
 
-If you treat rebasing as a way to clean up and work with commits before you push them, and if you only rebase commits that have never been available publicly, then you'll be fine.
+If you treat rebasing as a way to clean up and work with commits before anyone could have based work on them, either because they have not been available publicly or it is a branch that you are the only one supposed to work on, then you'll be fine.
```
ben: What about something more like this:

> If you only ever rebase commits that have never left your own computer, you'll be just fine.
>
> If you rebase commits that have been pushed, but that no one else has based commits from, you'll also be fine.

limdor: Sounds good to me, it clearly cover the two options, rebase of local commits and rebase of commits that you are the only one working on them.

PR #1155, opened 2019-01-09
single-quote normalization

ajax333221:
Found the use of this single quote ’ instead of ' (which looks slightly different).

I found them on files with extension:

- `.sketch` *
- `.pdf` *
- `.png` *
- `.ico` *
- `.asc` (15 files)

_* I didn't modify these since they are binary files which happen to contain them in their gibberishy binariness._

Then also noticed double single-quotation was being used (perhaps for some templating or idk), but I was careful not to mess with them (used a negative lookbehind regexp: `(?<!’)’` )


jnavila:
Not sure that changing apostrophes to quotes is correct, though.

ben:
Looks like [technically](https://asciidoctor.org/docs/user-manual/#curved) we should be using this syntax to be completely unambiguous:

```
"`Hello there,`" said Joe`'s sister.
```

According to that document, the single-quote character should always produce a dumb vertical single-quote. However, it looks like the PDF version is getting smart quotes anyways – [this section](https://github.com/progit/progit2/blob/master/book/01-introduction/sections/about-version-control.asc) contains the text `Many people's version-control`, which is rendered with a proper apostrophe in the [HTML output](https://git-scm.com/book/en/v2/Getting-Started-About-Version-Control) as well as the latest PDF download.

PR #1160, opened 2019-01-12
Added clarification about repo re-initialization

josh-byster:
Added a line about running `git init` even after cloning the repo, as it can be unclear that this command is non-destructive.


josh-byster:
Got it! Just changed. If you want, I can also move the tags to their original line so there's no diff on line `57`.

ben:
I think this is fine. The goal of the index tags is to get you to the right page for the thing you're looking for, and I doubt any reasonable typesetting system would break a page between these two sentences. Thanks!

PR #1164, opened 2019-01-19
Format quoted option string as <code> in HTML

katrinleinweber:
This suggestion aligns the formatting with the next paragraph. Checked for other occurrences of ``` `--\w+' ``` but found none.


katrinleinweber:
You're welcome, and thanks for merging :-)

PR #1165, opened 2019-01-22
create new branch stable

a-c-sreedhar-reddy:


ben:
Going to decline this one. In this scenario we're taking a dependency on the submodule, and assuming it already has a `stable` branch. I don't see that we're adding anything to the explanation by complicating it with a branch we just created.

PR #1175, opened 2019-02-14
Fix usage of sudo to edit file

rexkerr:
- vim should not be run as root, but instead use `sudo -e`
     (or sudoedit) which creates a temporary copy of the file
     to edit and then copies it over the original if you modify the
     contents.


rexkerr:
Using `sudo -e` also ensures that the users EDITOR is used rather than assuming vim.

ben:
I like it, thanks!

PR #1179, opened 2019-02-15
Undoing things: Clarify what "git checkout" does to local files

rpjday:
The current explanation of "git checkout -- <file>" seems a bit
vague, so expand on it ever so slightly.

### DIFF
--- a/book/02-git-basics/sections/undoing.asc
+++ b/book/02-git-basics/sections/undoing.asc
@@ -130,8 +130,8 @@ You can see that the changes have been reverted.
 [IMPORTANT]
 =====
 It's important to understand that `git checkout -- <file>` is a dangerous command.
-Any changes you made to that file are gone -- Git just copied another file over it.
-Don't ever use this command unless you absolutely know that you don't want the file.
+Any local changes you made to that file are gone -- Git just copied the most recently-committed version of that file over top of it.
+Don't ever use this command unless you absolutely know that you don't want those unsaved local changes.
 =====
 
 If you would like to keep the changes you've made to that file but still need to get it out of the way for now, we'll go over stashing and branching in <<ch03-git-branching#ch03-git-branching>>; these are generally better ways to go.


ben:
Nice, thanks!

bripmccann:
Maybe I'm being unnecessarily picky, because the meaning is clear, but '<i>over top of</i>' sounds odd to me. Certainly '<i>**on** top of</i>' is the more common idiom—but in the case of copying, the old wording, ['<i>copy over</i>', seems to be the most common](https://trends.google.com/trends/explore?q=copy%20over,copy%20over%20top,copy%20on%20top).

But like I said, the meaning is still clear. I don't find '<i>copy over</i>' vague, but if it does add clarity for others I'd say it's worth a little awkwardness.

(P.S. I don't know if people usually comment on pull requests after they're merged? My bad if I should've opened an issue or something.)

ben:
I'd agree that it isn't that common, and it definitely feels a bit odd to me saying it out loud, but it does get the point across, and I don't think anybody will misread that passage to mean something else.

Do you want to submit a PR so we can see what it looks like?

HonkingGoose:
Hi @ben, you've merged PR #1246, doesn't that close this issue? Or is there more work to be done here still?

aollier:
@HonkingGoose Which issue are you talking about?

HonkingGoose:
Hi @aollier,

- This issue was first about: https://github.com/progit/progit2/pull/1179#issue-253463012
- And then about: https://github.com/progit/progit2/pull/1179#issuecomment-468422175

It seems that both those things are now fixed, so this issue (#1179) can be closed?

aollier:
@HonkingGoose Good.
Reminder: if you add a new commit or open a new pull request in order to close an issue, the recommended way of doing it is by [closing issues using keywords](https://help.github.com/en/articles/closing-issues-using-keywords).

Furthermore, if you work on Windows and are using TortoiseGit, you can see the issue related to the commit if the commit contains one of the keywords followed by # followed by the issue ID thanks to #1127.

bripmccann:
My commit message for PR #1246 was ambiguous. You'd usually be right to assume I was referencing an issue, not a different pull request. I've edited the commit message to clarify.

PR #1181, opened 2019-02-16
Tagging: Small amount of grammatical cleanup, enhance some explanations

rpjday:
Among other things:

 - explain that pushing tags pushes both types
 - add second variation of how to delete remote tags

### DIFF
--- a/book/02-git-basics/sections/tagging.asc
+++ b/book/02-git-basics/sections/tagging.asc
@@ -2,27 +2,27 @@
 === Tagging
 
 (((tags)))
-Like most VCSs, Git has the ability to tag specific points in history as being important.
-Typically people use this functionality to mark release points (v1.0, and so on).
-In this section, you'll learn how to list the available tags, how to create new tags, and what the different types of tags are.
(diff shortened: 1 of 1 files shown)


book/02-git-basics/sections/tagging.asc
```diff
@@ -238,9 +246,18 @@ To /git@github.com:schacon/simplegit.git
...
+
+[source,console]
+----
+$ git push origin --delete <tagname>
+----
+
 ==== Checking out Tags
 
-If you want to view the versions of files a tag is pointing to, you can do a git checkout, though this puts your repository in ``detached HEAD'' state, which has some ill side effects:
+If you want to view the versions of files a tag is pointing to, you can do a `git checkout`, which puts your repository in ``detached HEAD'' state, which has some ill side effects:
```
ben: 🤔 I think there are too many `which`s here. If you don't like the "though this" structure, maybe this should be two separate sentences.

rpjday: OK, mostly reverted that change.

PR #1183, opened 2019-02-16
Revision selection: Couple minor clarifications and updates

rpjday:

### DIFF
--- a/book/07-git-tools/sections/revision-selection.asc
+++ b/book/07-git-tools/sections/revision-selection.asc
@@ -1,7 +1,7 @@
 [[_revision_selection]]
 === Revision Selection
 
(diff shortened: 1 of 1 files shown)


book/07-git-tools/sections/revision-selection.asc
```diff
@@ -1,7 +1,7 @@
 [[_revision_selection]]
 === Revision Selection
 
-Git allows you to refer to a set of commits or a range of commits in several ways.
+Git allows you to refer to a single commit, set of commits or range of commits in a number of ways.
```
ben: I might be misremembering, but I think the Oxford comma is in use everywhere else in the book. For consistency, could we use it here too?

rpjday:
OK, done.

PR #1192, opened 2019-02-25
Changed 'C5' to 'C6' in the rebase example.

Splarv:
Looked like by logic there must be 'C6', but not 'C5'.


ben:
I don't think that's true. Here's the figure from the "merge example":



And here's the rebase figure:



So the original text is correct:

> […] the snapshot pointed to by `C4'` is exactly the same as the one that was pointed to by `C5` in the merge example.

Splarv:
Interesting, but I see this figure:

Ah, I see. You mentioned Figure 28.

PR #1198, opened 2019-03-09
fetch and push with different repositories

aollier:
Added a section that explains how to fetch from one repository and to push to another one.

### DIFF
--- a/book/06-github/sections/2-contributing.asc
+++ b/book/06-github/sections/2-contributing.asc
@@ -43,6 +43,7 @@ Here's how it generally works:
 5. Open a Pull Request on GitHub.
 6. Discuss, and optionally continue committing.
 7. The project owner merges or closes the Pull Request.
+8. Put the changes back in your GitHub public repository.
 
 This is basically the Integration Manager workflow covered in <<ch05-distributed-git#_integration_manager>>, but instead of using email to communicate and review changes, teams use GitHub's web based tools.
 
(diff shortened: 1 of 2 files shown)


ben:
See https://github.com/progit/progit2/pull/1200/commits/9a48cdaf347c25174d188042ff919881e97cbd42 for my suggestions.

Splarv:
May be this is matter of taste, but I think that will be better mark an origin repository as "origin" and a fork repository as "fork". And so the default configuration for a common GitHub user will be as simple as:
git clone _origin_url_
git remote add **fork** _fork_url_
git config remote.pushDefault **fork**

ben:
Naming _is_ one of the hardest problems. 😁 

Naming remotes after their fork names scales to more than one fork, which is nice as a maintainer when you have three or four forks you pull from regularly.

Splarv:
May be there are different use cases, at least two: a common submitter without write access and common maintainer. And each use case may suggest different scheme or workflow which will be mostly convenient. I don't know what workflow is better for a maintainer. But I think that an example of the workflow of the common newbie submitter must be clearly mentioned, because it is a newbie who poorly understand how to use git for his purposes.

I think that my workflow for the newbie is much more natural and clear:

1. The original repository is called "original".
2. The fork repository is called "fork".
3. In this workflow the fetch (pull) is always from the original repository. No one need to pull from the own fork repository. And so is the default behaviour of an original repository in git. (Git fetch all branches from an original by default.)
4. But all pushes must always go to the own fork repository (to create a pull request) and so is for the remote.pushDefault option.

So is a common workflow for the newbie submitter will be:
_create fork repository on GitHub site_
git clone _origin_url_ _(and so all branches will be fetched from the origin repository by default)_
cd _directory_of_repository_
git remote add **fork** _fork_url_
git config remote.pushDefault **fork** _(and so all branches will be pushed to the fork by default)

_for each feature repeat:_
git checkout -b _my_feature_branch_ **origin**/_master (or another project depended mainstream branch, which is needed)_ 
_do some work, after commit:_
git pull --rebase
git push
_create pull request for the "_my_feature_branch_"

In case of the committed example:

_create fork repository on GitHub site_
git clone _fork_url_ _(the fork will have name origin and all branches will be fetched from the fork by default, IMHO useless)_
cd _directory_of_repository_
git remote add _origin_name_ _origin_url_ _(origin repository will be named origin_name)_
git config remote.pushDefault **origin** 

_for each feature repeat:_
git checkout -b _my_feature_branch_ _origin_name_/_master (or another project depended mainstream branch, which is needed)_ 
_do some work, after commit:_
git pull --rebase
git push
_create pull request for the "_my_feature_branch_"

PR #1209, opened 2019-03-18
Fix spelling error.

HonkingGoose:
I made a spelling error, I only saw this after you've merged it into master.... :worried: 

It's supposed to be ``include`` not ``included``. The other lines all use ``include``, so I was in the wrong here.


ben:
Oh no! Thanks for fixing!

Issue #1210, opened 2019-03-19
Change http links to https.


HonkingGoose:
I searched for all matches of the string **http://** in the progit2/master source. The result was 529 matches across 74 files.

Shouldn't we use the more secure https links instead? Is there a reason that there are still http links? Is this something that you want fixed?

ben:
This sounds like it would be a great change. We should be using SSL links wherever possible.

HonkingGoose:
@ben Would you like me to go through the lot and change them?

ben:
I don't want you to feel like you have to do this all at once. If you want to, maybe take it a chapter at a time? Or do a search/replace for a single hostname all at once? But if you've got time and want to, I'd love to review a PR that does it in one big patch.

HonkingGoose:
Let's suppose I want to change the lot. What would be the logical way to make that change?

Which one of these would be preferred:

---

### Commit often:
1. Make new branch.
1. Make a lot of commits for each change.
1. Run Travis CI.
1. Merge into master.

---

### Commit often, but squash at the end:
1. Make new branch.
1. Make a lot of commits for each change.
1. Squash all commits into one large commit. (This would need to be done by you, I don't yet know how to use git to rebase anything without getting my ``head`` detached :smile: .
1. Run Travis CI.
1. Merge into master.

---

### Commit once for all changes:
1. Make new branch.
1. Make one big commit for all changes.
1. Run Travis CI.
1. Merge into master.

---

### Commit once per chapter:
1. Make new branch per chapter.
1. Make one big commit for all changes in chapter.
1. Run Travis CI.
1. Merge into master.

---

For reference, this is my current level of git understanding:
I just use ``git add -p``, ``git commit -v``, and ``git push``, and ``git fetch`` and ``git merge`` in my current usage of git. So interactively rebasing things is a bridge too far, for now.

---

So if you can tell me which workflow (or your own better workflow) I should use, I can decide if I want to get this underway. :+1:

ben:
My personal style for this repo is "lots of commits, not necessarily focused." In short I just take an action that I might want to undo, then make a commit. Sometimes that means I just finished an hour of work and have to go do something else, sometimes that means I just ran `sed` and made a hundred automated fixes.

In short, we're not super pedantic here. Go ahead and open a PR, and start pushing commits however feels best to you, and ping me when you'd like feedback.

[...]

PR #1211, opened 2019-03-19
Chapter 1 http -> https

HonkingGoose:
@ben

Here's the http -> https fix for chapter 1.

- I noticed some links were redirecting to a new website, so I've changed those links.
- I've removed the trailing / at the end of the urls.

I've build the .html files locally from the .asc files, they looked good.

Fixes part of #1210


ben:
✨

rllane17 (approved):

PR #1212, opened 2019-03-19
Git version 2.19.1 has new expanded output for git add -h

HonkingGoose:
@ben Git 2.19.1 on Ubuntu 18.10 has more output when using ``git add -h``.

Is this something you want to merge into master? I don't know what version of git the book is based on?


ben:
We try to stay current. Our goal is that if a user sees different output on their system than in the book, it _should_ be because their Git is out of date. We don't always live up to that, though, so thanks for making it more true!

PR #1215, opened 2019-03-19
Chapter 4 http -> https.

HonkingGoose:
@ben

Onwards to Chapter 4, I'm skipping chapter 2 and 3, because I found no mention of http:// in there.

I only found one relevant link that points to a third party http source. So I've updated that link.

The other mentions were in the gitlab.asc and gitweb.asc files, and those links mention user owned hosted servers (Gitlab or GitWeb), so those can stay in http I think?


ben:
Seems reasonable to me.

PR #1219, opened 2019-03-19
Chapter 8 http -> https.

HonkingGoose:
@ben 

I'm skipping chapter 7, I found no http:// links there.

I found a http:// link at book/08-customizing-git/sections/attributes.asc:
>First, you'll need to install `docx2txt`; you can download it from http://docx2txt.sourceforge.net[].

I'm not sure if this should be replaced or not. I've found this: https://sourceforge.net/projects/docx2txt/ but I don't know if that is the proper replacement for it...


ben:
It looks like that is the correct replacement. Want to include it in this PR?

HonkingGoose:
I've replaced the link to ``docx2txt`` in the PR.

PR #1224, opened 2019-03-19
contributing.md has outdated link which redirects to new domain.

HonkingGoose:
This link is outdated, using it redirects to the new domain: https://www.sketchapp.com/

By the way, it appears Sketch is now on version 52, according to https://www.sketchapp.com/


ben:
There are other efforts to get the diagram source out of Sketch, so hopefully this won't be an issue for much longer. Thanks!

PR #1228, opened 2019-03-20
Fix some spelling and https issues for translating.md

HonkingGoose:
@ben I've fixed some spelling and https issues here.

While fixing the issues, I noticed that the style of the text is not in line with the rest of the text in the book and on the rest of the progit2 source. Maybe this needs a rewrite or editing pass?


ben:
Nice, thanks. I think most of this text was written by a non-native-English speaker, which accounts for much of the syntax, but the spelling mistakes were an easy fix. Thanks!

PR #1229, opened 2019-03-20
Major rewrite of TRANSLATING.md

HonkingGoose:
Hi @ben,

I took the liberty of starting a major rewrite of TRANSLATING.md

### Main changes:
1. Changed table to proper GitHub Flavoured Markdown format. (I had problems getting the table to behave after my edits, so I rewrote it in proper GFM).
1. Added new table with the three stages a translation can be in.
1. Rewrite of the Travis CI sections.
1. Removal of Beyond Git (I don't think a section on hacking on git should be in a file about translating.)
1. General cleanup of text.

### Minor changes:
1. Changed some header text.
1. Changed header level.

### Feedback wanted on:

Maybe the Travis section can be made clearer? Maybe you have some suggestions on this? You use the system daily I think... :smile: 

And of-course I would like some feedback on the style/content of this rewrite.

PR #1230, opened 2019-03-21
Reinstate Beyond Pro Git section.

HonkingGoose:
@jnavila @ben

I've reinstated the old section about "Beyond Pro Git" to the end of the file.

Sequence of events:
* Made a new branch based on master
* ``git checkout 039fe2a934ae19b7d0c45599fc8662c588de9b61``
* Opened translating.md
* Manually copy the section Beyond Pro Git
* Made a new branch to rollback based on master
* Manually pasted the Beyond Pro Git section into master/translating.md

I think/know that this way is probably not the best way of rolling things back (essentially I just manually undid the work).

Also I'm not sure what you're supposed to do once changes are merged into master, I know not to use git --rebase on things that you've pushed, because that annoys everybody, because you re-write history under their feet?

I'm not sure how this rollback could be done cleaner. There seem to be 3 versions of reset, and also you can use ``checkout **specific commit sha1 id**``, so I'm getting quite confused. :worried: 

So I'm also looking for a bit of advice on how to do this properly and not in a hacky quick-fix way... :smile: 

For reference: #1229

### DIFF
--- a/TRANSLATING.md
+++ b/TRANSLATING.md
@@ -88,3 +88,15 @@ Copy that file, and put it in your working directory. Commit the .yml file and p
 ## Setting up a publication chain for e-books
 
 This is a technical task, please ping @jnavila to get started with epub publication.
+
+## Beyond Pro Git
+
+Translating the book is the first step. Once this is finished, you could consider translating the user interface of Git itself.
+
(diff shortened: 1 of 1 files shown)


jnavila:
Thank you. This section was clearly a way to hire the sparse translators for related projects :unicorn:

PR #1231, opened 2019-03-22
Mac → macOS

### LINKED ISSUE #1197: Mac Operating System
#1161 did not replace all occurences of "Mac" refering to the Mac Operating System.

I think that "Mac" in the following lines should be replaced by "macOS":

book/01-introduction/sections/installing.asc:36:(((`Mac`, installing)))
book/01-introduction/sections/installing.asc:54:You can also install it as part of the GitHub for `Mac` install.
book/01-introduction/sections/installing.asc:56:You can download that tool from the GitHub for `Mac` website, at http://mac.github.com[]. (not sure for this one)
book/02-git-basics/sections/getting-a-repository.asc:21:for `Mac`:
book/04-git-server/sections/git-on-a-server.asc:8:Here we'll be demonstrating the commands and steps needed to do basic, simplified installations on a Linux-based server, though it's also possible to run these services on `Mac` or Windows servers.
book/08-customizing-git/sections/attributes.asc:20:For instance, Xcode projects on the `Mac` contain a file that ends in `.pbxproj`, which is basically a JSON (plain-text JavaScript data format) dataset written out to disk by the IDE, which records your build settings and so on.
book/08-customizing-git/sections/config.asc:248:We'll use path names in the examples that work on `Mac` and Linux systems; for Windows, you'll have to change `/usr/local/bin` to an executable path in your environment.
book/08-customizing-git/sections/config.asc:252:We'll use the `Mac` path for the executable; in other systems, it will be where your `p4merge` binary is installed.
book/08-customizing-git/sections/config.asc:391:This is because Windows uses both a carriage-return character and a linefeed character for newlines in its files, whereas `Mac` and Linux systems use only the linefeed character.
book/08-customizing-git/sections/config.asc:403:If you're on a Linux or `Mac` system that uses LF line endings, then you don't want Git to automatically convert them when you check out files; however, if a file with CRLF endings accidentally gets introduced, then you may want Git to fix it.
book/08-customizing-git/sections/config.asc:411:This setup should leave you with CRLF endings in Windows checkouts, but LF endings on `Mac` and Linux systems and in the repository.
book/A-git-in-other-environments/sections/guis.asc:69:==== GitHub for `Mac` and Windows
book/A-git-in-other-environments/sections/guis.asc:71:(((GitHub for `Mac`)))(((GitHub for Windows)))
book/A-git-in-other-environments/sections/guis.asc:76:.GitHub for `Mac`.
book/A-git-in-other-environments/sections/guis.asc:77:image::images/github_mac.png[GitHub for `Mac`.]
book/A-git-in-other-environments/sections/guis.asc:99:GitHub for Windows can be downloaded from https://windows.github.com[], and GitHub for `Mac` from https://mac.github.com[].

@easybe do you agree with those changes ?

aollier:
completes #1161 and fixes #1197

### DIFF
--- a/book/01-introduction/sections/installing.asc
+++ b/book/01-introduction/sections/installing.asc
@@ -33,7 +33,7 @@ For more options, there are instructions for installing on several different Uni
 
 ==== Installing on macOS
 
-(((Mac, installing)))
+(((macOS, installing)))
 There are several ways to install Git on a Mac.
 The easiest is probably to install the Xcode Command Line Tools.(((Xcode)))
 On Mavericks (10.9) or above you can do this simply by trying to run 'git' from the Terminal the very first time.
@@ -51,9 +51,9 @@ A macOS Git installer is maintained and available for download at the Git websit
 .Git macOS Installer.
 image::images/git-osx-installer.png[Git macOS installer.]
 
-You can also install it as part of the GitHub for Mac install.
+You can also install it as part of the GitHub for macOS install.
 Their GUI Git tool has an option to install command line tools as well.
-You can download that tool from the GitHub for Mac website, at https://desktop.github.com[].
+You can download that tool from the GitHub for macOS website, at https://desktop.github.com[].
 
 ==== Installing on Windows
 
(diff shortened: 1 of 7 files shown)


ben:
Nice!

Issue #468, opened 2015-12-13
Arabic Translation


salahzidi:
I opened this issue to start an arabic translation for this great book:
https://github.com/salahzidi/progit2

ben:
Ping @schacon

schacon:
Hey @salahzidi, it doesn't look like much is actually translated. Once you get a chapter or two done, let me know and I can turn it into an official project.  Thanks!

salahzidi:
OK @schacon, I'm working on it!

jaksz:
Hey, is this still of interest? I can contribute to the Arabic translation.

HonkingGoose:
@jnavila It seems this issue has run its course, and can be closed? 

This might be the "official" repository for the translation into Arabic? 
https://github.com/progit2-ar/progit2 
If so, it needs to be listed in our TRANSLATE.md file, so that we point others the right way.

Anyway, it doesn't seem that any version of the Arabic Translation is yet pulled into the git-scm source. Arabic has the ```ar``` ISO code, I see no link ending on ```ar``` in the translations list at git-scm. I've checked the linked repository above, and there is not much translated yet.

---

For all people on the thread, the current guidelines for contributing a translation can be found here:
https://github.com/progit/progit2/blob/master/TRANSLATING.md

jnavila:
Ditto

HonkingGoose:
@jnavila Do you want a pull-request for this one, so that it's mentioned in the TRANSLATING.md table? Or do you want to leave this one as-is?

jnavila:
Good!

PR #1240, opened 2019-04-01
Add Arabic translation.

HonkingGoose:
Hi @jnavila, this adds the Arabic translation to the table.

There doesn't seem to be all that much happening in that repository though. On the introduction only a part is translated, so this needs a lot more work before it's in a state that it can be pushed to the git-scm page. (https://github.com/progit2-ar/progit2/blob/master/book/01-introduction/1-introduction.asc)

Are you sure you actually want this listed?

Reference: #468


jnavila:
Let's add it anyway.

PR #1243, opened 2019-04-03
Remove ubuntu 14.04 and upstart.

HonkingGoose:
Ubuntu 14.04 reaches End of Life this month, upstart is also old. (See: https://www.ubuntu.com/about/release-cycle)

I'm not sure about the:
>On other systems, you may want to use `xinetd`, a script in your `sysvinit` system, or something else -- as long as you get that command daemonized and watched somehow.

section so I'm leaving that in.


ben:
👍

PR #1246, opened 2019-04-09
Reword sentence about losing local changes

bripmccann:
1. Avoid awkward idiom, as discussed in https://github.com/progit/progit2/pull/1179#issuecomment-468422175.
2. Fix pronoun: As it was written, 'it' would've referred to the subject of the sentence, 'changes'. Besides the singular-plural disagreement, it should've said the _file_ was replaced, not the changes.


ben:
Yes! Thank you.

Issue #1249, opened 2019-04-17
Chapter 05 Distributed Git - some points


udyzr:
**Which version of the book is affected?**  Version 2.1.146, 2019-04-15 pdf, source text, others not checked

**Describe the bug:**

Book is addressed to Git newbies. Also chapter 05 Distributed Git. The less book material student this sort has absolved the less points are obvious. In the beginning zero points are obvious. Thus it is critical to make precise word selection and accurate descriptions. Otherwise mental journey is put onto undesired trails. Way back to turn point might be costly.
There are few few points in chapter 5 Distributed Git not unambiguous enough. Lets review them. 
**Issue 1**
Contributing to a Project - Private Small Team, line number 159 - Description insinuates John had made two commits till that line's of text point of time. Please compare with elaborations shortly before first transcript for John's machine (line number 112).
So I propose to change to : Continuing with this example, shortly afterwards, John tries to push them to the same server:
Fixed form does not raise ambiguity.

**Steps to reproduce:**

**Expected behavior:**

**Screenshots:**

**Additional context:**

**Desktop:**

- Operating system:
- Browser/application:
- Browser/application version:

**Smartphone:**

- Device:
- OS:
- Browser/application:
- Browser/application version:

**E-book reader:**

- Device:
- Software Update:

udyzr:
**Issue 2**
Contributing to a Project - Private Small Team, source text line number 273 reads: _Jessica can merge either origin/master or issue54 first — they’re both upstream..._
Open question: is for _master_ branch the branch _issue54_ really an upstream branch? In chapter about branching in Git I learned upstream branch is a branch being tracked by tracking branch. Here however _master_ and _issue54_ are two branches in the same repository.

udyzr:
**Issue 3**
Contributing to a Project - Private Small Team, Figure 65 Simple multiple developer workflow
I miss for Jessica's timeline a loop accounting for her committing on _issu54_ branch. Commit series on issu54 branch then her intention to push them to upstream (on master) are the reasons why Jessica is merging.
So I propose to draw on Jessica's timeline two subsequent commit loops. The second loop following Jessica's push.

ben:
General notes: 

* It would be helpful when reviewing these kinds of suggestions to include links to the relevant section on the git-scm.com website, or link to the line number you're talking about.
* If you're going to submit multiple issues, please submit them as separate issues, it helps keep the discussions focused.

## Issue 1

[...]

PR #1251, opened 2019-05-10
Reduce redundancy in story description

udyzr:
Prior to this fix the description insinuates John had made two commits
in this process phase. Please compare with elaborations shortly
before first transcript for John's machine.

### DIFF
--- a/book/05-distributed-git/sections/contributing.asc
+++ b/book/05-distributed-git/sections/contributing.asc
@@ -156,7 +156,7 @@ The basic format is `<oldref>..<newref> fromref -> toref`, where `oldref` means 
 You'll see similar output like this below in the discussions, so having a basic idea of the meaning will help in understanding the various states of the repositories.
 More details are available in the documentation for https://git-scm.com/docs/git-push[git-push].
 
-Continuing with this example, shortly afterwards, John makes some changes, commits them to his local repository, and tries to push them to the same server:
+Continuing with this example, shortly afterwards, John tries to push them to the same server:
 
 [source,console]
 ----


ben:
The prose had just finished showing how Jessica makes changes, commits them, and pushes them (in painful detail). The line you're trying to change is where we switch back to John. We need to tell his part of the story quickly, and having just gone through all this detail with Jessica, we can shorthand it with John by just saying what he does.

But the change you're suggesting removes the fact that John makes changes at all, and it seems like that's important information. Can you explain a bit more why you think this change is necessary?

udyzr:
> But the change you're suggesting removes the fact that John makes changes at all, and it seems like 
that's important information. Can you explain a bit more why you think this change is necessary?

I am not sure this. With this phrase left in text as it is John edits then commits twice - I am not sure that is the intended scenario. Please compare with line 112 as stated in issue #1251. Both line 159 and 112 use present tense, so in my understanding John performs these two operations twice.

Kindly see also https://github.com/progit/progit2/issues/1249#issuecomment-484567185.

udyzr:
Cit.
The first developer, **John**, clones the repository, **makes a change, and commits locally**. (The protocol messages have been replaced with …​ in these examples to shorten them somewhat.)
\< here about what Jessica makes \>
Continuing with this example, **shortly afterwards, John makes some changes, commits them to his local repository**, and tries to push them....

book/05-distributed-git/sections/contributing.asc
```diff
@@ -156,7 +156,7 @@ The basic format is `<oldref>..<newref> fromref -> toref`, where `oldref` means
 You'll see similar output like this below in the discussions, so having a basic idea of the meaning will help in understanding the various states of the repositories.
 More details are available in the documentation for https://git-scm.com/docs/git-push[git-push].
 
-Continuing with this example, shortly afterwards, John makes some changes, commits them to his local repository, and tries to push them to the same server:
+Continuing with this example, shortly afterwards, John tries to push them to the same server:
```
ben: Let's reword this sentence thus:

> Continuing with our example, John then tries to push his local changes to the same server:

udyzr:
> Let's also change line 128 to say something like this:
> 
> > While John is busy making his changes, the second developer (Jessica) does the same thing -- clones the repository and commits a change:

Actually I am happy with book's every piece of text else I do not complain about. Change this type would rather form a side-effect for this issue "Reduce redundancy in story description". Kindly consider an issue for this change - shouldn't it need a good argumentation as every issue else?

book/05-distributed-git/sections/contributing.asc
udyzr: Well, however myself wonders why phrase ", shortly afterwards, John ... tries" was able to work before change in PR but it isn't with PR.

ben:
> Change this type would rather form a side-effect for this issue "Reduce redundancy in story description".

I appreciate that you're trying to keep this PR focused, but I'm looking at the broader context of how we make this section more readable and useful. Your original suggestion is definitely a part of that, but I think we can do more, and this is the most on-topic place to do it.

> Kindly consider an issue for this change - shouldn't it need a good argumentation as every issue else - as well myself doen't share the need of change this kind?

I don't think that every sentence needs a lengthy discussion. That's actually my role here, to make these quick decisions that probably no one has a strong opinion about. If you don't agree with this suggestion that's completely okay, I'm happy to accept your contribution the way you'd like it to look.

> Furthermore I believe also in the end of the day it does not matter if Jessica and John are doing their changes parallel or subsequently in time.

For me it helps to visualize two people working on the code at the same time, and one of them pushes first. This doesn't matter on a technical level, but since this is in the part of the book where we're just introducing multi-person workflows, it helps to be as friendly as possible.

> Well, however myself wonders why phrase ", shortly afterwards, John ... tries" was able to work before change in PR but it isn't with PR.

I actually think it _doesn't_ work very well.

udyzr:
Kindly ask person who sees the need for additional change as pointed to put it into repository. Myself either still didn't review text from this perspective or does not see the need for this adaption. I can raise PR for changes myself is convinced to be needed. Also if proposed additional change should receive some critics in future I am not ready to defend it cause I don't understand its need. I appreciate your understanding.

> For me it helps to visualize two people working on the code at the same time
Even though. They two might have it (cloning, editing then committing) done literally at the same time or each after another one in very short time distance. This however does not matter. It is the order in which  their's two pushes are made which matters imho.
Should "While John is busy making his changes, the second developer" get some critics in future I am neither able nor ready to defend it. Please understand.

book/05-distributed-git/sections/contributing.asc
udyzr: From current point of view I am neither able nor ready to defend this proposal if one day it should need that.

ben: OK, let's stick to just your suggestions then. This isn't quite grammatically correct, let's change it to this:

> Continuing with this example, shortly afterwards, John tries to push his changes to the same server:

book/05-distributed-git/sections/contributing.asc
```diff
@@ -270,8 +270,10 @@ Switched to branch 'master'
 Your branch is behind 'origin/master' by 2 commits, and can be fast-forwarded.
 ----
 
-Jessica can merge either `origin/master` or `issue54` first -- they're both upstream, so the order doesn't matter.
-The end snapshot should be identical no matter which order she chooses; only the history will be different.
+Jessica can merge either `origin/master` or `issue54` first -- the order doesn't matter.
+The end snapshot will be identical no matter which order she chooses; only the history will be different.
+However let's mention the fact in this case to be worth to merge `issue54` first as it will be just fast-forward -
+no new commit is created - resulting in version tree bit more clearer.
```
ben: More grammatical updates:

> In this case, she chooses to merge `issue54` first, since it will be a fast-forward merge, which will result in a simpler history.

kaream671:
> Prior to this fix the description insinuates John had made two commits
> in this process phase. Please compare with elaborations shortly
> before first transcript for John's machine.

Kaream671

lex2491 (approved):

PR #1252, opened 2019-05-10
Fix: Inspecting commits made on topic branch since...

udyzr:
common ancestor with main branch.

Issue detected and fixed by consulting git-revisions manual page
(used git version is 2.21.0):
 <rev1>...<rev2> - Include commits that are reachable from either <rev1> or <rev2>
 but exclude those that are reachable from both...
 <rev1>..<rev2> - Include commits that are reachable from <rev2>
 but exclude those that are reachable from <rev1>...

### DIFF
--- a/book/05-distributed-git/sections/maintaining.asc
+++ b/book/05-distributed-git/sections/maintaining.asc
@@ -282,7 +282,7 @@ or, more concisely:
 $ git diff $(git merge-base contrib master)
 ----
 
-However, neither of those is particularly convenient, so Git provides another shorthand for doing the same thing: the ellipsis (double-dot) syntax.
+However, neither of those is particularly convenient, so Git provides another shorthand for doing the same thing: the double-dot syntax.
 In the context of the `git diff` command, you can put two periods after another branch to do a `diff` between the last commit of the branch you're on and its common ancestor with another branch:
 
 [source,console]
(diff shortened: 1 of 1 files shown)


ben:
From https://git-scm.com/docs/git-rev-parse#_dotted_range_notations:

> ### The .. (two-dot) Range Notation
> The ^r1 r2 set operation appears so often that there is a shorthand for it. When you have two commits r1 and r2 (named according to the syntax explained in SPECIFYING REVISIONS above), you can ask for commits that are reachable from r2 excluding those that are reachable from r1 by ^r1 r2 and it can be written as r1..r2.
> 
> ### The …​ (three-dot) Symmetric Difference Notation
> A similar notation r1...r2 is called symmetric difference of r1 and r2 and is defined as r1 r2 --not $(git merge-base --all r1 r2). It is the set of commits that are reachable from either one of r1 (left side) or r2 (right side) but not from both.

You're completely right! I can't believe we've been giving the wrong advice this whole time. Thanks!

ben:
Actually wait. Let me think through this. I need to go back and really understand these operators before I merge or reject this.

One thing I _do_ know is that an [ellipsis](https://en.wikipedia.org/wiki/Ellipsis) in modern typography is three dots.

udyzr:
You are right, I am wrong with using the term 'ellipsis' in my patch. Thanks and, sorry for that. Me will update my pr.

ben:
👍

alfredmyers:
> You're completely right! I can't believe we've been giving the wrong advice this whole time. Thanks!

Actually, the text was correct. Please see #1265 .

PR #1258, opened 2019-06-01
Add extra notes about format for a commit object

SiarheiBobryk:
From my point of view we should note here that a commit object
contains its parent commit(s)


ben:
Nice catch, thanks!

PR #1260, opened 2019-06-02
Extend the git commands list to get the manpage

SiarheiBobryk:
Recently I found the extra way to get the manpage so I'd like add it
to the book.


ben:
Yes, good addition. Thanks!

SiarheiBobryk:
👍

PR #1261, opened 2019-06-04
Note about moving the branch after a commit

SiarheiBobryk:
Actually the branch pointer moves after a commit, but not the HEAD in
this case. The HEAD still points to the branch.

So the branch pointer has moved only.

### DIFF
--- a/book/03-git-branching/sections/nutshell.asc
+++ b/book/03-git-branching/sections/nutshell.asc
@@ -113,7 +113,7 @@ $ vim test.rb
 $ git commit -a -m 'made a change'
 ----
 
-.The HEAD branch moves forward when a commit is made
+.The testing branch moves forward when a commit is made, The HEAD still points to the testing branch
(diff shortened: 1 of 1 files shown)


ben:
Let me give you a bit of motivation here. At this point in the book we're just trying to introduce the concept of branches, and the "HEAD branch" to a reader at this level is "the branch I have checked out." This is a less-pedantically-correct but also less-confusing version of the truth. We don't want to overwhelm readers who are completely new to Git with tons of technical detail, it'll drive them away.

Given all that, do you think this extra detail, while more correct, is important to readers in the section where we're just introducing the concept of branches?

SiarheiBobryk:
Sorry Ben. I did not think about it from this perspective. It makes sense for readers like me who started from Git internals chapter and then continue to read from the beginning of the book.
So yes, I agree with you. Let's do not overwhelm readers.

PR #1263, opened 2019-06-06
Simplify the way to get own branch from remote

SiarheiBobryk:
I'm not sure about older versions of Git, but my local Git version
(2.21.0) supports this way to automatically create local branch from
the remote one and checkout to it at the same time.

So I'd like to suggest this way to do that.


SiarheiBobryk:
Sorry. This way is described couple paragraphs below in the book.

PR #1264, opened 2019-06-12
Fix type of target branch when merging with git pull

Dobiasd:


book/03-git-branching/sections/remote-branches.asc
```diff
@@ -215,7 +215,7 @@ $ git fetch --all; git branch -vv
 While the `git fetch` command will fetch all the changes on the server that you don't have yet, it will not modify your working directory at all.
 It will simply get the data for you and let you merge it yourself.
 However, there is a command called `git pull` which is essentially a `git fetch` immediately followed by a `git merge` in most cases.
-If you have a tracking branch set up as demonstrated in the last section, either by explicitly setting it or by having it created for you by the `clone` or `checkout` commands, `git pull` will look up what server and branch your current branch is tracking, fetch from that server and then try to merge in that remote branch.
+If you have a tracking branch set up as demonstrated in the last section, either by explicitly setting it or by having it created for you by the `clone` or `checkout` commands, `git pull` will look up what server and branch your current branch is tracking, fetch from that server and then try to merge in that local branch.
```
ben: I think this is incorrect. Some terminology:

* **Remote branch**: a local ref that represents the state of a branch in a remote repository (like `origin/master`)
* **Remote-tracking branch**: a local branch (like `master`) that's set up to track a remote branch

 What's really happening in a pull where your HEAD is a remote-tracking branch is:

1. `git fetch origin/master`, which updates `origin/master` in the local repo.
1. `git merge origin/master`, which updates `master`. 

So the original wording is more correct here: you're "merging `origin/master` into your head". Does that make sense? I can see how it's confusing, but I think we've managed to at least be consistent in applying these two terms.

Dobiasd: Oh, damn. I think I misread
`try to merge in that [...] branch`
(semantically) as
`try to merge into the [...] branch`.

Of course, you are totally right. Thanks for the nice explanation, and sorry for opening this unnecessary pull request.

PR #1267, opened 2019-06-15
Fix the double quotes and comma typo

SiarheiBobryk:
I'm not an expert in English grammar, but I think here is the typo.

### DIFF
--- a/book/05-distributed-git/sections/contributing.asc
+++ b/book/05-distributed-git/sections/contributing.asc
@@ -103,7 +103,7 @@ In short, do as we say, not as we do.
 
 (((contributing, private small team)))
 The simplest setup you're likely to encounter is a private project with one or two other developers.
-``Private,'' in this context, means closed-source -- not accessible to the outside world.
+"Private", in this context, means closed-source -- not accessible to the outside world.
 You and the other developers all have push access to the repository.
 
 In this environment, you can follow a workflow similar to what you might do when using Subversion or another centralized system.


book/05-distributed-git/sections/contributing.asc
```diff
@@ -103,7 +103,7 @@ In short, do as we say, not as we do.
 
 (((contributing, private small team)))
 The simplest setup you're likely to encounter is a private project with one or two other developers.
-``Private,'' in this context, means closed-source -- not accessible to the outside world.
+"Private", in this context, means closed-source -- not accessible to the outside world.
```
ben: Ha, I know it looks wrong, but it's actually correct.

First, the quote style is an Asciidoctor-ism. From the [manual](https://asciidoctor.org/docs/user-manual/#curved):

> Single and double quotation marks are not rendered as curved quotation marks (also known as smart, curly, typographic or book quotation marks). When entered using the ' and " key, Asciidoctor outputs straight (dumb, vertical, typewriter) quotation marks.

We care about typography, so we made the change to the double-backtick and double-single-quote method early on in the second edition. It looks like the syntax has changed slightly in the last few years, and we should probably be using the new-style ``"`curve-direction mark`"`` syntax, but that's a big change to make.

Secondly, commas almost always go inside quotation marks. I'll break this rule if the quoted text is the subject of the sentence (_click "Edit", then…_), but that's not the case here.

Does that make sense?

SiarheiBobryk: yep! absolutely. 👍

PR #1272, opened 2019-06-29
Specify the branch the HEAD points to on init

SiarheiBobryk:
I would like to propose to specify the exact branch the HEAD points
to after the git init command invoking.


ben:
Nice, I like it.

SiarheiBobryk:
👍

PR #1273, opened 2019-07-07
Change the order of the three states of files.

ChenSL86:
"1.3 Getting Started - What is Git?" - "The Three States", the orders of "modified", "staged" and "committed" were mixed and I think it might be better to be all the same natural order.


ben:
Yeah, nice. I like it. Thanks!

PR #1281, opened 2019-07-25
Add "go to the root directory" step before cloning

italicize:


ben:
I'm going to decline this one. We have to assume a basic knowledge of the command-line environment, and selecting directories should be part of that. Also, even a complete beginner at the CLI will end up cloning something to their home directory, which is just fine for that stage of learning.

PR #1270, opened 2019-06-19
Aggregate core.editor command list in Appendix C

katrinleinweber:
Close #1269.

I have not tested most these commands, except for Atom & Sublime and am not sure about:

- [ ] list `gvim` at all, given that it's also a default setting: https://git-scm.com/docs/git-var

Maybe worth discussing, besides reviewing the commands:

- [ ] add "(default)" behind the ["Vim" mention in `first-time-setup.asc`](https://github.com/progit/progit2/blame/f52df9b858ee7533f15fd48248858881f6588105/book/01-introduction/sections/first-time-setup.asc#L83); There is no example command, and novices might not find https://git-scm.com/docs/git-var to clarify this by themselves.

### Preview

image

PR #1283, opened 2019-08-02
Expand core.editor command list in Appendix

katrinleinweber:
Following #1269 & #1270, it was suggested to me to include improve the table in these two ways, see commits & diff. Do you think that makes sense?


ben:
Nice. Thanks!

katrinleinweber:
You're welcome, and thanks for merging :-)

katrinleinweber:
Ah, oh! I wanted to include https://github.com/progit/progit2/pull/1284 here, but you were too fast :-D

ben:
Whoops! Sometimes I can't tell if a PR is ready for the merge button or not. If you include a checklist and not everything is checked I'll definitely hold off though.

PR #1295, opened 2019-08-24
SVG images checked and fixed (some redesigned)

max123kl:
font-weight altered to "700"
text positions and color fixed
some files redesigned to look as "former" PNG files

test files deleted

I created an [HTML table](http://zgastro.bplaced.net/file-list.html) that allows a straight comparison of SVG and PNG files.
I noticed some minor issues, which I want to fix as soon as possible.
The files have to be inspected closely


max123kl:
> Looks like these arrows got turned around:

Yes, I didn't notice. In general, I discovered that all arrowheads were converted incorrectly. In many cases it is not necessary to change this, because it is not relevant for the content of the drawing. It is only an appearance problem.
clean.svg fixed.

> It also looks like some of the refs got resized in a weird way.

I've noticed that. In some cases I had already corrected that.
I will check the diagrams again and touch them up one by one. But it will take some time to fix all the files.

max123kl:
As far as I can see, all the files are ready to merge.
If there is still an issue, please notify me.

ben:
😻

PR #1301, opened 2019-09-08
Update remotes.asc

max123kl:
Quotations fixed (lines 14 + 15, 121 +122)

### DIFF
--- a/book/02-git-basics/sections/remotes.asc
+++ b/book/02-git-basics/sections/remotes.asc
@@ -11,8 +11,8 @@ In this section, we'll cover some of these remote-management skills.
 [NOTE]
 .Remote repositories can be on your local machine.
 ====
-It is entirely possible that you can be working with a ``remote'' repository that is, in fact, on the same host you are.
-The word ``remote'' does not necessarily imply that the repository is somewhere else on the network or Internet, only that it is elsewhere.
+It is entirely possible that you can be working with a `remote` repository that is, in fact, on the same host you are.
+The word "remote" does not necessarily imply that the repository is somewhere else on the network or Internet, only that it is elsewhere.
 Working with such a remote repository would still involve all the standard pushing, pulling and fetching operations as with any other remote.
 ====
 
@@ -118,7 +118,7 @@ $ git fetch <remote>
 The command goes out to that remote project and pulls down all the data from that remote project that you don't have yet.
 After you do this, you should have references to all the branches from that remote, which you can merge in or inspect at any time.
 
-If you clone a repository, the command automatically adds that remote repository under the name ``origin''.
+If you clone a repository, the command automatically adds that remote repository under the name `origin`.
 So, `git fetch origin` fetches any new work that has been pushed to that server since you cloned (or last fetched from) it.
 It's important to note that the `git fetch` command only downloads the data to your local repository -- it doesn't automatically merge it with any of your work or modify what you're currently working on.
 You have to merge it manually into your work when you're ready.


ben:
No, the quotation here is intentional. In Asciidoc, this is the rule:

```asciidoc
"these are straight quotes"
``these are curly quotes''
```

Markdown does "smart quotes" and tries to guess what shape of quotation mark you want, but Asciidoc requires you to manually apply curvature for a certain direction.

Your patch downgrades the quotes to "straight quotes," which is not a look I'd want in a PDF or future print version.

max123kl:
I understand and support your intentions absolutely.
As a graphic designer with a university degree, it is very important for me to use the typographically correct punctuation and quotation marks.

You enter a string and let Asciidoctor render.
Entering the characters in your way is tedious and has a high error rate.
Since nowadays almost all computers use UTF-8 code, it is better to enter the desired character directly via keyboard shortcut.
Probably you are using a MacOS machine so here are the shortcuts:

`“`       left double quotes, "high 66" (English: opening)    —   opt/alt (+) 2
`”`	right double quotes, "high 99" (English: closing)   —   shift (+) opt/alt (+) 2

Quotation marks_macOS
https://www.key-shortcut.com/mac-osx/sonderzeichen-mac
Here you can find a lot of other shortcuts for special characters (also Windows and HTML)

ben:
Aha I see what you did. If this were a manuscript just for me, I'd be with you 100%. In fact, I was using the native UTF-8 en-dash character (`–`) while writing this, and we later decided that it was better to use the ASCII character sequence (`--`) that renders to the same result instead. 

In the end, it's much easier for new contributors to type the standard ASCII characters, regardless of their keyboard layout, OS, or editor, so we should stick to either the old (` ``quote'' `) or new (`` "`quote`" ``) Asciidoc notation.

max123kl:
Actually, the problem is that certain characters, easily accessible via the keyboard, are interpreted differently as control code by certain editors (such as Markdown and Asciidoctor).

On the other hand, my workaround is to bypass the control code and enter as many characters as possible in the output (all defined by UTF-8 code) via keyboard shortcuts. Because these characters are rarely used as control code, I don't leave the rendering of the characters to the editors.

All these characters can be entered directly with any keyboard. Differences in how this can be done are in the keyboard layout used and the operating system of the computer.

As the knowledge of how to use this is very basic for many users in the technical professions,, I have created a [file](https://github.com/progit/progit2-de/blob/master/Special_Characters.asc) to help translators/authors to use the typographically correct spelling for the new [German repo](https://github.com/progit/progit2-de).

However, even for English texts you can quickly define a few basic rules that can prevent an output like this ([taken from](https://github.com/progit/progit2/blob/master/book/03-git-branching/sections/remote-branches.asc)):

From_remote-branches1

From_remote-branches2

From_remote-branches3

From_remote-branches4

OK, let me first mention the different dashes and then move on to the quotation marks and describe how to use them correctly.

There are three Dashes of different widths.

"`-`" The hyphen should only be used for word separations at the end of a line and for word compositions. Sadly it is often used incorrectly.

"`–`" The en-dash is used in German texts, among other uses, to connect parts of a sentence visually and contentwise. 

"`—`" The em-dash is used in English texts where the "en-dash" is used in German texts. With AsciiDoctor you achieve this correctly with `--`.

Quotation marks are used very differently in different languages. The following characters are common for English books:

The straight quotation mark (single ' or double ") should only be used in sample texts for code.

It is better and more correct to use the curved quotation marks. Unicode (UTF-8) defines more than 20 [different characters](http://www.amp-what.com/unicode/search/quote).
The following are common in English texts:

"`“`" U+201c left double quotation mark

"`”`" U+201d right double quotation mark
 
"`‘`" U+2018 left single quotation mark 
 
"`’`" U+2019 right single quotation mark

PR #1306, opened 2019-09-26
bash prompt example: add a missing escape

dktapps:
Without this escape, the `__git_ps1` is evaluated directly when `.bashrc` is run, and not when a new prompt is needed, which causes the prompt to never update.


dktapps:
my mistake, I didn't notice I was using double quotes, the example uses single quotes ...

PR #1313, opened 2019-10-14
Fix #1310

### LINKED ISSUE #1310: Add commands for migration from SVN to Git on Windows Systems
v2: https://git-scm.com/book/en/v2/Git-and-Other-Systems-Migrating-to-Git

**Describe the bug:**
As a Windows user, I cannot use the suggested command to create the authors mapping list for migrating from SVN to Git.

**Steps to reproduce:**
https://git-scm.com/book/en/v2/Git-and-Other-Systems-Migrating-to-Git says I could use `svn log --xml --quiet | grep author | sort -u | \
  perl -pe 's/.*>(.*?)<.*/$1 = /'` to create a list of authors to pass into Git for subsequent migration from SVN.
As I don't have access to a Linux PC, I can't use that command on my machine.

**Expected behavior:**
There should be an alternative command for the major platforms Windows, Linux (already there) and MacOS (probably the same or very similar to the Linux command) using as many default Windows tools as possible while using as few third party programs as possible.

**Desktop:**
- Operating system: Windows 10

CodingSpiderFox:
Fixes #1310

### DIFF
--- a/book/09-git-and-other-scms/sections/import-svn.asc
+++ b/book/09-git-and-other-scms/sections/import-svn.asc
@@ -26,7 +26,9 @@ $ svn log --xml --quiet | grep author | sort -u | \
   perl -pe 's/.*>(.*?)<.*/$1 = /'
 ----
 
-Windows users can find advice and a sample command for creating an authors list here: https://docs.microsoft.com/en-us/azure/devops/repos/git/perform-migration-from-svn-to-git?view=azure-devops
+[NOTE]
+If you're trying this on a Windows machine, this is the point where you'll run into trouble.
+Microsoft have provided some good advice and samples at https://docs.microsoft.com/en-us/azure/devops/repos/git/perform-migration-from-svn-to-git
 
 That generates the log output in XML format, then keeps only the lines with author information, discards duplicates, strips out the XML tags.
 (Obviously this only works on a machine with `grep`, `sort`, and `perl` installed.)
--- a/book/09-git-and-other-scms/sections/import-svn.asc
+++ b/book/09-git-and-other-scms/sections/import-svn.asc
@@ -29,7 +29,7 @@ $ svn log --xml --quiet | grep author | sort -u | \
 [NOTE]
 ====
 If you're trying this on a Windows machine, this is the point where you'll run into trouble.
-Microsoft have provided some good advice and samples at https://docs.microsoft.com/en-us/azure/devops/repos/git/perform-migration-from-svn-to-git
+Microsoft have provided some good advice and samples at https://docs.microsoft.com/en-us/azure/devops/repos/git/perform-migration-from-svn-to-git[].
 ====
 
 That generates the log output in XML format, then keeps only the lines with author information, discards duplicates, strips out the XML tags.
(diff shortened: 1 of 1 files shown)


book/09-git-and-other-scms/sections/import-svn.asc
```diff
@@ -26,6 +26,8 @@ $ svn log --xml --quiet | grep author | sort -u | \
   perl -pe 's/.*>(.*?)<.*/$1 = /'
 ----
 
+Windows users can find advice and a sample command for creating an authors list here: https://docs.microsoft.com/en-us/azure/devops/repos/git/perform-migration-from-svn-to-git?view=azure-devops
```
ben: 🤔 Could you enclose this in a `[NOTE]`, and change the text to be something like this?

```suggestion
If you're trying this on a Windows machine, this is the point where you'll run into trouble.
Microsoft have provided some good advice and samples at https://docs.microsoft.com/en-us/azure/devops/repos/git/perform-migration-from-svn-to-git[].
```

book/09-git-and-other-scms/sections/import-svn.asc
```diff
@@ -26,6 +26,10 @@ $ svn log --xml --quiet | grep author | sort -u | \
   perl -pe 's/.*>(.*?)<.*/$1 = /'
 ----
 
+[NOTE]
+If you're trying this on a Windows machine, this is the point where you'll run into trouble.
+Microsoft have provided some good advice and samples at https://docs.microsoft.com/en-us/azure/devops/repos/git/perform-migration-from-svn-to-git
+
```
ben: In order to keep our consistency, let's add the delimiters here too:

```suggestion
[NOTE]
====
If you're trying this on a Windows machine, this is the point where you'll run into trouble.
Microsoft have provided some good advice and samples at https://docs.microsoft.com/en-us/azure/devops/repos/git/perform-migration-from-svn-to-git
====

```

book/09-git-and-other-scms/sections/import-svn.asc
```diff
@@ -26,6 +26,15 @@ $ svn log --xml --quiet | grep author | sort -u | \
...
 ----
 
+[NOTE]
+If you're trying this on a Windows machine, this is the point where you'll run into trouble.
+Microsoft have provided some good advice and samples at https://docs.microsoft.com/en-us/azure/devops/repos/git/perform-migration-from-svn-to-git
+[NOTE]
+====
+If you're trying this on a Windows machine, this is the point where you'll run into trouble.
+Microsoft have provided some good advice and samples at https://docs.microsoft.com/en-us/azure/devops/repos/git/perform-migration-from-svn-to-git
+====
```
ben: Whoops, looks like accepting that change doubled this up. ☹️

book/09-git-and-other-scms/sections/import-svn.asc
```diff
@@ -26,6 +26,12 @@ $ svn log --xml --quiet | grep author | sort -u | \
   perl -pe 's/.*>(.*?)<.*/$1 = /'
 ----
 
+[NOTE]
+====
+If you're trying this on a Windows machine, this is the point where you'll run into trouble.
+Microsoft have provided some good advice and samples at https://docs.microsoft.com/en-us/azure/devops/repos/git/perform-migration-from-svn-to-git
```
ben: Ooh, one more recommendation: punctuation and auto-linking:

```suggestion
Microsoft have provided some good advice and samples at https://docs.microsoft.com/en-us/azure/devops/repos/git/perform-migration-from-svn-to-git[].
```

ben:
Nice, thanks!

PR #1314, opened 2019-10-14
Getting Help: fix inconsistency with '-h' option

phil-blain:
Only the short `-h` option shows the concise help output; the long `--help` option actually opens the manpage, as mentioned at the beginning of the "Getting Help" section. 

This PR fixes the inconsistency in the text.


ben:
Nice! I ❤️ PRs that _remove_ text.

PR #1317, opened 2019-10-17
Update import-svn.asc

CodingSpiderFox:
Move paragraph directly under the code snippet which it describes
Also remove unnecessary braces


ben:
Yes, better. Thanks!

PR #1320, opened 2019-10-23
Fix cd command for windows

mixxamm:
'/users/' was missing.


ben:
Are you running this under WSL? IIRC the path we have is correct for MinGW Git.

mixxamm:
Yup, my bad.

PR #1321, opened 2019-10-25
Add tip about abbreviating long git options

### LINKED ISSUE #1315: Mention abbreviating long options
I just learned that any long options (`--option`) in Git can be abbreviated (ex. `--opt`, `--o`) as long as it's resolvable (see https://git-scm.com/docs/gitcli#_abbreviating_long_options). I did a search and I don't think this is mentionned in the book. I think it would be valuable to mention that (some options are quite long!), though I'm not sure where best it would fit (maybe at the beggining of Appendix C ?)

HonkingGoose:
This pull-request changes:

- Add tip about abbreviating long git options to the beginning of Appendix C.

Fixes #1315


ben:
Nice. Thanks!

Issue #1304, opened 2019-09-23
EOL declared for git-tf Tool


max123kl:
If you follow that [link](https://gittf.codeplex.com) from the file 
/book/09-git-and-other-scms/sections/client-tfs.asc, 
you will learn that this tool has reached the EOL status.
Is there any intention to change the text in the book or place a note on the status there?

PR #1322, opened 2019-10-25
Add caution git-tf EOL, change links to Codeplex

HonkingGoose:
This pull-request changes:

- Change links to codeplex to point to the archival pages.
- Add caution about git-tf being End-of-Life.

I think this section needs more work though, I would advocate to remove the git-tf part wholesale.
Git-tf clone from codeplex.com will also not work anymore I think.

But a big removal like that is up to @ben and team to decide. 😄 

Related: #1304


ben:
Created #1326 to track the big removal. Thanks!

PR #1352, opened 2019-12-28
Add caution: use `git filter-repo` instead of `git filter-branch`

HonkingGoose:
This commit adds:
- a caution block in the relevant part of the book.
- a link to the repository/documentation.

Context:
The git-scm reference documentation for `filter-branch`
warns against using `git filter-branch`,
and recommends using `git filter-repo`.

### DIFF
--- a/book/07-git-tools/sections/rewriting-history.asc
+++ b/book/07-git-tools/sections/rewriting-history.asc
@@ -304,6 +304,13 @@ The command is `filter-branch`, and it can rewrite huge swaths of your history, 
 However, it can be very useful.
 You'll learn a few of the common uses so you can get an idea of some of the things it's capable of.
 
+[CAUTION]
+====
+`git filter-branch` has many pitfalls, and is no longer the recommended way to rewrite history.
+Instead, consider using `git-filter-repo`, which is a Python script that does a better job for most applications where you would normally turn to `filter-branch`.
+Its documentation and source code can be found at https://github.com/newren/git-filter-repo[].
+====
+
 [[_removing_file_every_commit]]
 ===== Removing a File from Every Commit
 


C-git-commands.asc
```diff
@@ -568,6 +568,13 @@ In <<ch07-git-tools#_removing_file_every_commit>> we explain the command and exp
 
 In <<ch09-git-and-other-systems#_git_p4>> and <<ch09-git-and-other-systems#_git_tfs>> we use it to fix up imported external repositories.
 
+[CAUTION]
+====
+`git filter-branch` is no longer the recommended way to rewrite history.
+Instead use the Python script `git filter-repo`, this script is safer and more performant.
+This script is not included with git, you will need to download and install the script yourself.
+Read the documentation for `git filter-repo` [here](https://github.com/newren/git-filter-repo/).
+====
```
ben: Let's drop this diff. This section is only listing places where a reader can find mentions of specific commands, kind of like an index, and it's not the right place for this kind of warning.

book/03-git-branching/sections/rebasing.asc
```diff
@@ -231,6 +231,14 @@ The opposing point of view is that the commit history is the *story of how your
...
 This is the camp that uses tools like `rebase` and `filter-branch` to tell the story in the way that's best for future readers.
 
+[CAUTION]
+====
+`git filter-branch` is no longer the recommended way to rewrite history.
+Instead use the Python script `git filter-repo`, this script is safer and more performant.
+This script is not included with git, you will need to download and install the script yourself.
+Read the documentation for `git filter-repo` [here](https://github.com/newren/git-filter-repo/).
+====
+
```
ben: Same comment as above, this isn't the right place to get into this kind of detail.

book/07-git-tools/sections/rewriting-history.asc
```diff
@@ -304,6 +304,14 @@ The command is `filter-branch`, and it can rewrite huge swaths of your history,
...
 You'll learn a few of the common uses so you can get an idea of some of the things it's capable of.
 
+[CAUTION]
+====
+`git filter-branch` is no longer the recommended way to rewrite history.
+Instead use the Python script `git filter-repo`, this script is safer and more performant.
+This script is not included with git, you will need to download and install the script yourself.
+Read the documentation for `git filter-repo` [here](https://github.com/newren/git-filter-repo/).
+====
+
```
ben: I like the idea here, but let's reword it a bit:

```asciidoc
`git filter-branch` has many pitfalls, and is no longer the recommended way to rewrite history.
Instead, consider using `git-filter-repo`, which is a Python script that does a better job for most applications where you would normally turn to `filter-branch`.
Its documentation and source code can be found at https://github.com/newren/git-filter-repo[].
```

HonkingGoose:
I've incorporated your feedback, and force pushed the branch, so please check again if this is what you want. 😄

ben:
🎯

PR #1360, opened 2020-01-02
Use “their” instead of “his”

hidde:
So that no gender is implied in this sentence and the explanation makes sense for people of all genders.

(I did not open an issue as the contributing guidelines request, as the change is very small)

### DIFF
--- a/book/05-distributed-git/sections/distributed-workflows.asc
+++ b/book/05-distributed-git/sections/distributed-workflows.asc
@@ -67,7 +67,7 @@ This is a variant of a multiple-repository workflow.
 It's generally used by huge projects with hundreds of collaborators; one famous example is the Linux kernel.
 Various integration managers are in charge of certain parts of the repository; they're called _lieutenants_.
 All the lieutenants have one integration manager known as the benevolent dictator.
-The benevolent dictator pushes from his directory to a reference repository from which all the collaborators need to pull.
+The benevolent dictator pushes from their directory to a reference repository from which all the collaborators need to pull.
 The process works like this (see <<wfdiag_c>>):
 
 1.  Regular developers work on their topic branch and rebase their work on top of `master`.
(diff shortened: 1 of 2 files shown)


lazarljubenovic:
Is it somehow a problem that the sentence refers to a male collaborator? Maybe the author imagined a male person, or himself? 

> the explanation makes sense for people of all genders

The explanation makes perfect sense with any gender. Nobody will think "oh no the collaborator cannot be a woman because this text said _his_, I must find another version-control system which women can use too".

hidde:
> Is it somehow a problem that the sentence refers to a male collaborator? Maybe the author imagined a male person, or himself?

I think there's no problem referring to a male collaborator, but when doing so using 'they' (which includes all), it avoids reinforcing the stereotype that programmers are males.

See also: [Nonsexist language](https://geekfeminism.wikia.org/wiki/Nonsexist_language) (TLDR: please take requests like this one seriously)

ben:
We tried to make our gendering at least evenly distributed, and it looks like we succeeded in that; there are exactly four examples with a [male character](https://github.com/progit/progit2/search?utf8=%E2%9C%93&q=his&type=) and four with a [female](https://github.com/progit/progit2/search?utf8=%E2%9C%93&q=her&type=). Of course, this was in 2014, before I knew as much about gender as I do now. 

I'd totally merge a PR that improves the text in this way. Probably when the person being referred to doesn't have a name, we should use a they/them pronoun, and perhaps we change one of our named characters to a they/them as well? @hidde do you want to do that here?

hidde:
@ben Many thanks for context, was not aware these characters were distributed 50/5, that's great!  Change one of the named characters to _they_ and use _they_ / _them_ whenever a character is unnamed. Should I start a new PR for that?

ben:
> Should I start a new PR for that?

No need for that, you can just push more commits to this branch and we'll review them all at once.

hidde:
I found one more instance of an unnamed character, and have [changed “his” to “their”](https://github.com/progit/progit2/pull/1360/commits/d630140f7fe5451575b24b1c6fd20b95e756d755). Given there are only two named characters, I think we may just leave them gendered for now?

@ben From my side, this PR is ready for review.

ben:
I like it, this is definitely a good step. Thanks!

PR #1362, opened 2020-01-07
Use monosace font for 'master' branch name

Morganov:


lazarljubenovic:
Back when I read the book, I had a feeling that some instances of "master branch" were supposed be adjectives, not precise name of the branch. Not sure if that was intended.

C-git-commands.asc
```diff
@@ -95,7 +95,7 @@ To take a directory and turn it into a new Git repository so you can start versi
 
 We first introduce this in <<ch02-git-basics-chapter#_getting_a_repo>>, where we show creating a brand new repository to start working with.
 
-We talk briefly about how you can change the default branch from ``master'' in <<ch03-git-branching#_remote_branches>>.
+We talk briefly about how you can change the default branch from `master` in <<ch03-git-branching#_remote_branches>>.
```
ben: This change is interesting, and I don't have a clear idea of which way is better.

* Monospace indicates that `master` is a technical term you would type into a terminal, which it is, but
* Double quotes indicate that this is a string whose value you can change, which it is

I'm leaning towards someting like maybe this, but only a little. Do you have a strong feeling?

```suggestion
We talk briefly about how you can change the default branch name from ``master`` in <<ch03-git-branching#_remote_branches>>.
```

Morganov: I had a doubt about it. Now, I agree with your suggestion: the word "master" should be wrapped by double quotes if the word "name" will be added.

ben:
🛳 Thanks!

PR #1370, opened 2020-01-31
Merge pull request #1370 from UgmaDevelopment/patch-1

UgmaDevelopment:
Just a quick fix: the Users on a Windows system are located in a folder named Users, not directly on the C drive

### DIFF
--- a/book/02-git-basics/sections/getting-a-repository.asc
+++ b/book/02-git-basics/sections/getting-a-repository.asc
@@ -26,7 +26,7 @@ $ cd /Users/user/my_project
 for Windows:
 [source,console]
 ----
-$ cd /C/Users/user/my_project
+$ cd C:/Users/user/my_project
 ----
 
 and type:
--- a/book/02-git-basics/sections/getting-a-repository.asc
+++ b/book/02-git-basics/sections/getting-a-repository.asc
@@ -26,7 +26,7 @@ $ cd /Users/user/my_project
 for Windows:
 [source,console]
 ----
-$ cd /c/user/my_project
+$ cd /C/Users/user/my_project
 ----
 
 and type:


Morganov:
I assume, it would be better to use `~` symbol instead. 

It works fine in `git-bash` on Windows 10.

UgmaDevelopment:
Sure! I was just matching the convention of the other lines.

Note that Windows users will not recognize `~` as the denoting the home
folder; that's a *nix convention.

Maybe a short sentence mentioning that the symbol represents the current
user's home folder would make that unmistakably clear.

ben:
Pretty sure that line was written back in the MinGW days, when the bash shell location actually was `/c/user`.

Now the question is this: what's the most common shell readers will use to interact with Git in 2020, and what is the path to the home directory in that shell? Is it cmd.exe? PowerShell? LSW bash?

UgmaDevelopment:
It looks like `cd C:/Users/user/my_project` works in CMD, Powershell, and Git Bash, whereas `cd /C/Users/user/my_project` only works in Git Bash (and likely other non-windows shells)

|            | cd ~/my_project | cd /C/Users /user/my_project | cd C:/Users/user/my_project |
|------------|-----------------|------------------------------|-----------------------------|
| CMD        | X               | X                            | W7, W10                     |
| Powershell | W7, W10         | X                            | W7, W10                     |
| Git Bash   | W7, W10         | W7, W10                      | W7, W10                     |

I like the one that works in all of them.

book/02-git-basics/sections/getting-a-repository.asc
```diff
@@ -26,7 +26,7 @@ $ cd /Users/user/my_project
 for Windows:
 [source,console]
 ----
-$ cd /c/user/my_project
+$ cd /C/Users/user/my_project
```
ben: In light of your lovely research, should we update this like so?

```suggestion
$ cd C:/Users/user/my_project
```

ben:
✨ Thanks!

PR #1378, opened 2020-02-22
Include html file to release

Morganov:


ben:
I like it. Does that HTML file display properly, or does it rely on CSS/JS that's in other files?

Morganov:
Hm.. This file has all css/js embedded in it.
It has a little bit different view in comparison with git-scm.com, but it's
acceptable to read.
So you can easily review the result by building the book manually - it's a
part of build result.

ben:
:+1:

PR #1381, opened 2020-02-24
Replace mention of "ticket" with "issue"

UgmaDevelopment:
I suspect it's clearer to talk about issues instead of tickets. Everyone knows what an issue is. Not everyone has worked with a ticket system; maybe they'll wonder if tickets are part of git. Sticking to the word "issues" seems to allow us to lose no clarity while reducing the chance for confusion.

I'm willing to consider other opinions, however.


ben:
Yeah, this agrees better with the branch name too. Nice!

PR #1382, opened 2020-02-24
Adding an Asciidoc quick reference

UgmaDevelopment:
Hopefully this will help anyone else that is unfamiliar with AsciiDoc, like I was.

While I was at it, I also changed the capitalization to AsciiDoc, which appears to be the official stylized capitalization. (https://www.google.com/search?q=asciidoc)

### DIFF
--- a/README.asc
+++ b/README.asc
@@ -7,7 +7,7 @@ You can find this book online at: https://git-scm.com/book
 Like the first edition, the second edition of Pro Git is open source under a Creative Commons license.
 
 A couple of things have changed since open sourcing the first edition.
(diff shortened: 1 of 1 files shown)


README.asc
```diff
@@ -7,7 +7,7 @@ You can find this book online at: https://git-scm.com/book
 Like the first edition, the second edition of Pro Git is open source under a Creative Commons license.
 
 A couple of things have changed since open sourcing the first edition.
-For one, we've moved from Markdown to the amazing Asciidoc format for the text of the book.
+For one, we've moved from Markdown to the amazing AsciiDoc format for the text of the book; here's an https://powerman.name/doc/asciidoc[AsciiDoc quick reference].
```
ben: Let's use this one instead: https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/

UgmaDevelopment: Sounds good!

PR #1383, opened 2020-02-24
NFS clarification

UgmaDevelopment:
I was just thinking that adding a link, explaining what NFS stands for, would be helpful.


ben:
Yup, I like it.

PR #1384, opened 2020-02-25
A missing word and a clarification

UgmaDevelopment:
As it was, the parts of the sentence didn't quite agree with each other. Adding "HTTP" alongside "HTTPS" restored the agreement that I suspect the sentence originally had.

Also, I changed "these ports" to "their ports" because the specific ports that the protocols use aren't mentioned. "These ports" seems to refer to ports that were previously mentioned in the text; "their ports" intends to refer to the ports belonging to the protocols.


ben:
Nice catch, thanks!

PR #1387, opened 2020-02-29
Update NOTE about showing history

Morganov:


Morganov:
I'm not sure that is the correct place for whole this note. Possibly, it would be much better to move it to "02-git-basics/sections/viewing-history.asc" file.
@ben what do you think?

PR #1373, opened 2020-02-06
Switch to asciidoctor 2.0.10

jnavila:
Change style to new specifications.

reference to #1355

PR #1388, opened 2020-02-29
Bump asciidoctor-epub3 to 1.5.0.alpha.14

slonopotamus:
This PR doesn't depend on #1373.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,11 +6,11 @@ gem 'asciidoctor', '1.5.6.2'
 gem 'json'
 gem 'awesome_print'
 
-gem 'asciidoctor-epub3', '~> 1.5.0.alpha.9'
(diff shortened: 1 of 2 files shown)


slonopotamus:
There's an interesting story about your negative leveloffset. Asciidoctor has an [rule about appendix](https://asciidoctor.org/docs/user-manual/#user-appendix):

> the first subsection of the appendix must be a level-3 section, not a level-2 section. That’s because an appendix cannot contain chapters.

When you're applying -1 leveloffset, your appendix subsections become level-2 sections so they're *not* subsections of appendix anymore, but instead are standalone chapters.

slonopotamus:
I suddenly understood that your Gemfile uses latest `1.5.0.alpha.WHATEVER` asciidoctor-epub3 version.

slonopotamus:
But you still need to fix leveloffset.

PR #1389, opened 2020-03-01
Stop adding negative leveloffset for EPUB

slonopotamus:
asciidoctor-epub3 1.5.0-alpha-14 and newer now parses whole book as a single document.

Negative leveloffset
1. Breaks appendix hierarchy because Asciidoctor doesn't allow appendix subsections to be higher that level 3. When leveloffset is applied to them, they become level 2 and are no longer children of appendix and instead become standalone chapters
2. Transforms progit into a multi-part book instead of multi-chapter book, what I think is wrong

After removal of negative offset, EPUB is structurally equivalent to other formats.

### DIFF
--- a/progit.asc
+++ b/progit.asc
@@ -8,8 +8,6 @@ Scott Chacon; Ben Straub
 :front-cover-image: image:book/cover.png[width=1050,height=1600]
 :icons: font
 
-ifdef::ebook-format[:leveloffset: -1]
-
 include::book/license.asc[]
 
 include::book/preface_schacon.asc[]


slonopotamus:
@ben ping? Are there any concerns that prevent this PR from being merged?

ben:
I'd love to. @jnavila do you have an idea about what effect this would have? Why did we add this setting in the first place?

jnavila:
This setting was needed for proper part slicing of the book in epub format. IIRC, otherwise, the whole book structure would be included in one file, and cross links would not work.

Now, @slonopotamus is in charge of asciidoctor-epub3, so we can trust his judgement on the capabilities of newer versions. Do we need anything else? Maybe a version condition in Gemfile?

slonopotamus:
Your Gemfile accepts all asciidoctor-epub3 1.5.x versions. As I already said, major rework happened in alpha.14 and only minor fixes were done since then. So you _could_ change lower bound to 1.5.0.alpha.14 to ensure that users get better results.

HonkingGoose:
Hi @slonopotamus! :wave: 

As you might have seen, we've migrated from `Asciidoctor 1.5.6.2` to `Asciidoctor 2.0.11`.

We've also upgraded from `'asciidoctor-epub3', '~> 1.5.0.alpha.9'` to `'asciidoctor-epub3', '1.5.0.alpha.18'`

Can you pull in the current `progit2/master` branch into this branch and see if this change is still needed?

Thanks! :smile: :+1:

slonopotamus:
TOC without patch:

image


TOC with patch:

image

jnavila:
I have to check that our script on git-scm does not break on this.

jnavila:
Sorry, I'm dumb. This is only for epub format and we use raw html format for the script.

PR #1390, opened 2020-03-03
Remove outdated information.

OlingCat:
GitHub Free now includes unlimited private repositories.


OlingCat:
> It's not correct. There is a limitation for private repositories under "Free" plan.
> Please, review https://github.com/pricing page for details.

Sorry, I should investigate that before make this PR.

PR #1391, opened 2020-03-03
Include images into HTML file

Morganov:
This parameter required to create a single portable HTML file during the build ([source](https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#images))

Release script already fixed [here](https://github.com/progit/progit2-pub/pull/6)


ben:
Nice, thanks!

PR #1401, opened 2020-03-23
Clarify the non-incremental nature of tree objects.

eric-s-raymond:
The one confirming detail left out of the exposition.

### DIFF
--- a/book/10-git-internals/sections/objects.asc
+++ b/book/10-git-internals/sections/objects.asc
@@ -344,6 +344,10 @@ If you follow all the internal pointers, you get an object graph something like 
 .All the reachable objects in your Git directory.
 image::images/data-model-3.png[All the reachable objects in your Git directory.]
 
+To be clear clear, the tree object built for a commit references *all* of the files and directories visible at the commit
+It does not only reference the content altered from the parent commit.
+If the tree object for a commit fails to include a file that was present in its parent commit, that absence is interpreted as deletion of the file.
+
 ==== Object Storage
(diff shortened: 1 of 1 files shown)


book/10-git-internals/sections/objects.asc
```diff
@@ -344,6 +344,10 @@ If you follow all the internal pointers, you get an object graph something like
 .All the reachable objects in your Git directory.
 image::images/data-model-3.png[All the reachable objects in your Git directory.]
 
+To be clear clear, the tree object built for a commit references *all* of the files and directories visible at the commit
+It does not only reference the content altered from the parent commit.
+If the tree object for a commit fails to include a file that was present in its parent commit, that absence is interpreted as deletion of the file.
+
```
ben: Oh yeah, I like the intent here, there's definitely information that was left unspecified.

Let's reword a bit. How about this?

```suggestion
To be clear, a commit's tree structure represents the complete contents of the working directory at that moment in time.
If the parent commit's tree includes a file that is not present in this commit's tree, that absence is interpreted as deletion of the file.

```

What do you think?

ugultopu:
```suggestion
To be clear, a commit's tree structure represents the complete contents of the staging area at that moment in time.
If the parent commit's tree includes a file that is not present in this commit's tree, that absence is interpreted as deletion of the file.

```

revmarcuspeter-cyber (approved):

PR #1405, opened 2020-03-25
Harmonize slashes as backward all Windows paths

katrinleinweber:
I'm not entirely sure about this. Maybe `\` everywhere would be less confusing for Windows users?

However, both the Git Bash and cmd.exe seem to accept both versions, at least on my Win10.

### DIFF
--- a/C-git-commands.asc
+++ b/C-git-commands.asc
@@ -67,13 +67,13 @@ Accompanying the configuration instructions in <<ch01-getting-started#_editor>>,
 |Textpad (Windows 64-bit) |`git config --global core.editor "'C:/Program Files/TextPad 5/TextPad.exe' -m` (Also see note below)
 |Vim |`git config --global core.editor "vim"`
 |Visual Studio Code and VSCodium |`git config --global core.editor "code --wait"`
-|WordPad |`git config --global core.editor '"C:\Program Files\Windows NT\Accessories\wordpad.exe"'"`
+|WordPad |`git config --global core.editor '"C:/Program Files/Windows NT/Accessories/wordpad.exe"'"`
 |Xi | `git config --global core.editor "xi --wait"`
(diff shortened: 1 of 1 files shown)


ben:
Yeah, I'm not sure about this either. We've got to make sure this works with cmd.exe, powershell, msysgit-bash, and WSL, and probably all the way back to XP. Can anyone verify that the backslash works in all those places?

max123kl:
I can confirm that **both** (slash and backslash) work with "cmd.exe", "Powershell" and "MINGW64" (Win7/64bit)
Perhaps a supplementary note declaring both versions as acceptable could clarify possible confusion for Windows users?

katrinleinweber:
Thanks for testing :-)

> […] note declaring both versions as acceptable

IMHO an argument to switch to backslashes to _avoid_ confusion in the first place ;-)

ben:
Lovely. Thanks to you both!

katrinleinweber:
You're welcome and thanks for the review @max123kl :-)

max123kl:
@katrinleinweber  
proofreader wanted
As I can see in your profile, you are staying in Germany and perhaps you are German-speaking too?
I hope that I can convince you to review some articles in the [German Progit2-Repo](https://github.com/progit/progit2-de) from time to time.
Along with my co-maintainer and another 2 collaborators we have almost completed the German translation of this repo.
However, it would be a great help to us if the articles were read again by someone independently in order to correct any technically incorrect phrases.
Greetings from Kaiserslautern

katrinleinweber:
I had indeed considered it, but I already feel guilty for not keeping the [two](https://crowdin.com/profile/katrinleinweber) [other](https://github.com/katrinleinweber/opensource.guide/tree/de) similar translation projects as up-to-date as I technically could :-/

PR #1406, opened 2020-03-25
Clarify meaning of "VS Code"

katrinleinweber:
Since both versions use the same CLI call, I think it would be good SEO to spell out both their official names here ;-)


ben:
❤️

PR #1407, opened 2020-03-26
replaced 'Working Directory' with 'Working Tree', as per general terminology

enikao:
The image at https://git-scm.com/book/en/v2/images/areas.png labels the first box in the upper left "Working Directory". However, all over the text, including the image caption at https://git-scm.com/book/en/v2/Getting-Started-What-is-Git%3F, the term "Working tree" is used.

I don't know how to regenerate the png, any hints?


ben:
The PNGs were generated from an older version of Sketch, which doesn't make them very easy to update. That's why we're moving to SVGs, and I think your patch here will be better once we're finished with that transition. Thanks!

PR #1411, opened 2020-03-31
Move VSCodium and use the proper command

HonkingGoose:
VSCodium uses `codium` as as the name of the binary. Using `git config --global core.editor "code --wait"` will not work properly with VSCodium.

This pull-request moves VSCodium to it's own line, and uses the proper command to setup VSCodium as the core editor.

Reference for VSCodium terminal usage:
https://github.com/VSCodium/vscodium/blob/master/DOCS.md#how-do-i-open-vscodium-from-the-terminal


HonkingGoose:
Thanks for merging! 😄

PR #1413, opened 2020-04-07
Error when describing the Smart Protocol?

mildsunrise:
It currently says (emphasis mine):

> Each line starts with a 4-character hex value specifying **how long the rest of the line is**.
> Your first line starts with 00a5, which is hexadecimal for 165, meaning that 165 bytes remain on that line.

This is inconsistent with the examples; the length seems to include its own 4 characters, and the linefeed:

~~~ js
line = '003fe2409a098dc3e53539a9028a94b6224db9d6a6b6 refs/heads/master\n'
line.length === 0x3f
~~~

### DIFF
--- a/book/10-git-internals/sections/transfer-protocols.asc
+++ b/book/10-git-internals/sections/transfer-protocols.asc
@@ -153,9 +153,9 @@ $ ssh -x git@server "git-receive-pack 'simplegit-progit.git'"
 The `git-receive-pack` command immediately responds with one line for each reference it currently has – in this case, just the `master` branch and its SHA-1.
 The first line also has a list of the server's capabilities (here, `report-status`, `delete-refs`, and some others, including the client identifier).
 
-Each line starts with a 4-character hex value specifying how long the whole the line is (including the 4-character length itself and the trailing linefeed).
-Your first line starts with 00a5, which is hexadecimal for 165, meaning the line is 165 bytes long.
-The next line is 0000, meaning the server is done with its references listing.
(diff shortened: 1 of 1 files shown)


mildsunrise:
Also modified the text to say 'chunk' rather than 'line'. This is more accurate, as a chunk can have multiple lines of data, and may not have a trailing linefeed.

book/10-git-internals/sections/transfer-protocols.asc
```diff
@@ -153,9 +153,9 @@ $ ssh -x git@server "git-receive-pack 'simplegit-progit.git'"
 The `git-receive-pack` command immediately responds with one line for each reference it currently has – in this case, just the `master` branch and its SHA-1.
 The first line also has a list of the server's capabilities (here, `report-status`, `delete-refs`, and some others, including the client identifier).
 
-Each line starts with a 4-character hex value specifying how long the rest of the line is.
-Your first line starts with 00a5, which is hexadecimal for 165, meaning that 165 bytes remain on that line.
-The next line is 0000, meaning the server is done with its references listing.
+The data is transmitted in chunks. Each chunk starts with a 4-character hex value specifying how long the chunk is (including the 4 bytes of the length itself). Chunks usually contain a single line of data and a trailing linefeed.
```
ben, line 156: Could we stick to the one-sentence-per-source-line convention here? It makes future diffs much easier to read.

mildsunrise:
Resolved, thanks for pointing out!

ben:
:sparkles: thanks!

PR #1415, opened 2020-04-13
Fix sentence flow: not "however", but "so"

pbodnar:
Rationale: "However" would start a contradiction, while there is a simple implication, so "so" or "therefore" needs to be used.

Hopefully I didn't misunderstand the text and the suggested correction is OK... :)


book/10-git-internals/sections/maintenance.asc
```diff
@@ -50,7 +50,7 @@ cac0cab538b970a37ea1e769cbbde608743bc96d refs/tags/v1.0
 
 If you update a reference, Git doesn't edit this file but instead writes a new file to `refs/heads`.
 To get the appropriate SHA-1 for a given reference, Git checks for that reference in the `refs` directory and then checks the `packed-refs` file as a fallback.
-However, if you can't find a reference in the `refs` directory, it's probably in your `packed-refs` file.
+So, if you can't find a reference in the `refs` directory, it's probably in your `packed-refs` file.
```
ben, line 53: I love this, it's a lot better. Just one request for flow:

```suggestion
So if you can't find a reference in the `refs` directory, it's probably in your `packed-refs` file.
```

PR #1420, opened 2020-04-21
submodules.asc: clarify DbConnector/try-merge example

salewski:
Fix AsciiDoc `source,console` block to render separately the consecutive commands `git branch try-merge c771610` and `git merge try-merge`. This is accomplished by separating the commands with an empty line, and by omitting the unneeded "(DbConnector) " prefix in front of the `git merge...` command prompt. Puts the example into the form consistent with the other command blocks, and also enhances readability.

State explicitly in the prose that "try-merge" will be the name of the branch.


ben:
Oh yeah, that's probably some leftover copy-paste junk. Thanks for improving this!

PR #1430, opened 2020-05-17
Typo error correction

storytellerr:
a typo error correction.


SpicyCarrot1550:
> You should have to do these things only once on any given computer; they'll stick around between upgrades.

This sentence is grammatically correct as it is.
There is no typo.

> You should have to do these things only once on any given computer

and

> [They will] stick around between upgrades

are both independently complete sentences.  

Therefore, they should be joined with either a conjunction (“, as” in this case) or a semicolon.

ben:
Since this is mostly a question of style, I'm going to stick with the original text. If you think this is a significant improvement, I'm willing to have my mind changed. 😄

PR #1435, opened 2020-05-22
Add link to Martin Fowlers git patterns page

### LINKED ISSUE #1434: Include link in book to Martin Fowlers page on branching patterns
I think the book could be improved by comparing the different workflows that are possible with git, and highlighting the benefits and trade-offs that each flow has. However this might be considered "scope/feature-creep"... Luckily somebody else already made a good resource on this, so that we don't have to duplicate the effort! 😄

This website explains the different git branching patterns that you can choose from, and the pros/cons of each: https://martinfowler.com/articles/branching-patterns.html

The page is still being worked on by Martin Fowler, but what's already there is of high quality and very helpful. Quote from bottom of linked page:
>I'm releasing this article in installments. I've now finished describing all the patterns on my todo list. Next week I'll briefly talk about how they tie into the popular branching strategies gitflow, github flow, and trunk-based development.

I suggest adding a link to Martin Fowlers website in the https://github.com/progit/progit2/blob/master/book/03-git-branching/sections/workflows.asc section of the book.

If this is something that the ProGit teams want, I'll open a pull-request with the suggested change for review. 👍

HonkingGoose:
Closes #1434

### DIFF
--- a/book/05-distributed-git/sections/distributed-workflows.asc
+++ b/book/05-distributed-git/sections/distributed-workflows.asc
@@ -83,6 +83,18 @@ image::images/benevolent-dictator.png[Benevolent dictator workflow.]
 This kind of workflow isn't common, but can be useful in very big projects, or in highly hierarchical environments.
 It allows the project leader (the dictator) to delegate much of the work and collect large subsets of code at multiple points before integrating them.
 
+[[_patterns_for_managing_source_code_branches]]
+==== Patterns for Managing Source Code Branches
+
+[NOTE]
+====
+Martin Fowler has made a guide "Patterns for Managing Source Code Branches".
+This guide covers all the common Git workflows, and explains how/when to use them.
+There's also a section comparing high and low integration frequencies.
+
+https://martinfowler.com/articles/branching-patterns.html
+====
+
 ==== Workflows Summary
 
 These are some commonly used workflows that are possible with a distributed system like Git, but you can see that many variations are possible to suit your particular real-world workflow.


HonkingGoose:
Hi @ben, here's your pull-request! 🚀 

- I tried to give a overview of why the page is worth reading, but it might be too verbose for the book.
- I also offered my advice on at least reading the Integration Frequency section.

I'm not too sure what's the best place for the link, up top, or after the explanation.

So I'm eagerly awaiting your feedback, and your recommended changes. 😄

ben:
Yes! This fits in pretty well in this section of the book. I have two requests:

1. Can we make this shorter, and structure it as a `[[NOTE]]`? It's not _our_ content, so we should call it out with formatting as a pointer to something else.
1. Let's drop the list of topics. It will inevitably get out of date, and having to follow that project in order to keep this one correct is a chore I don't think anyone wants.

HonkingGoose:
I've made the changes you wanted.

However I also noticed that we already have a section of the book with some of the workflows:
https://github.com/progit/progit2/blob/master/book/05-distributed-git/sections/distributed-workflows.asc

Do you still want the link to be where it would be now, or would you rather move it to just before the summary of distributed workflows section of the book?

I've marked the pull-request as a draft, so that we only merge it if we're happy with it. 😄

ben:
Oh, hmm. Now that you mention it, this content really does belong in chapter 5. Chapter 3's section is meant to introduce the idea of branches, but mostly in a repo-local context. The multi-repo model is in chapter 5, and I think that's where these workflows belong, I think. Can we relocate this note to the [summary area](https://github.com/progit/progit2/blob/master/book/05-distributed-git/sections/distributed-workflows.asc#L86) of that section in chapter 5?

HonkingGoose:
@ben I moved the text to just before the summary in chapter 5. I also trimmed the text a bit.

ben:
Great, thanks!

PR #1446, opened 2020-06-28
Use code highlighting

cbachhuber:
@mo-gul proposed this [while working on the german translation](https://github.com/progit/progit2-de/pull/110). I agree that using code highlighting would be nice in this sentence. What do you think?


ben:
Yeah, this seems appropriate here. Thanks!

PR #1450, opened 2020-07-01
Change dead link within section on git within Visual Studio

HonkingGoose:
The link that's changed in this commit is dead.

I went to: https://docs.microsoft.com/en-us/visualstudio/windows/?view=vs-2019
to see where the docs for using git with the current version of Visual Studio 2019 can be found.
The site has a section with tasks, when I click on Version control, it takes me to:
https://docs.microsoft.com/en-us/azure/devops/repos/?view=azure-devops

On the azure dev-ops page they have a section called **Get started with Git or GitHub**. That section contains information about using git with Visual studio or from the command line. 
Those sections look like this:
https://docs.microsoft.com/en-us/azure/devops/repos/git/gitquickstart?view=azure-devops&tabs=visual-studio

I'm not sure what the best way to deal with this is, or what the best link should be. I've just pointed the link to the basic landing page after clicking on **Get started with Git or Github** from the Visual Studio 2019 docs.

Note that changing the link is not enough to get the section in order though: #1451

### DIFF
--- a/book/A-git-in-other-environments/sections/visualstudio.asc
+++ b/book/A-git-in-other-environments/sections/visualstudio.asc
@@ -22,4 +22,4 @@ image::images/vs-2.png[The ``Home'' view for a Git repository in Visual Studio]
 
 Visual Studio now has a powerful task-focused UI for Git.
 It includes a linear history view, a diff viewer, remote commands, and many other capabilities.
-For complete documentation of this feature (which doesn't fit here), go to http://msdn.microsoft.com/en-us/library/hh850437.aspx[].
+For more on using Git within Visual Studio go to: https://docs.microsoft.com/en-us/azure/devops/repos/git/command-prompt?view=azure-devops[].


HonkingGoose:
Hi @ben, I did my best with this, but I'm not sure on how to move forward with this pull-request, we could fix the link, but that's only fixing a symptom, and not the cause...

I welcome your feedback and input on this pull-request and the linked issue (#1451)! :smile:

ben:
Yeah this is a sticky wicket. I think I'd prefer to hold off until we can do a more extensive rewrite. 

In any case, the link you're providing leads to a page entitled "Azure Repos Git Documentation." I don't see a single item there that shows how to use Git in Visual Studio, so it's not the right place for this link to point to. A quick Google doesn't turn up anything definitive either.

Can you find a better link? If so I'll merge this, otherwise we should close it out in favor of a bigger rework.

HonkingGoose:
> Yeah this is a sticky wicket. I think I'd prefer to hold off until we can do a more extensive rewrite.

Yeah I see this is a bit of a sticky wicket.

The "Azure Repos Git Documentation" does have `git` items in their sidebar though:
Sidebar -> Tutorials -> Commit, share and sync your code
https://docs.microsoft.com/en-us/azure/devops/repos/git/gitquickstart?view=azure-devops&tabs=visual-studio
This link only has screens for Visual Studio 2017 though, I don't know if the screens for Visual Studio 2019 are different...

I agree it's not really clear where the proper documentation for Visual Studio 2019 and git lives...
It's strange that there is no clear entry point into the Visual Studio 2019 git integration docs that I can find... 😞

HonkingGoose:
Can you find a better link? If so I'll merge this, otherwise we should close it out in favor of a bigger rework.

Hi @ben, I think I found a better link. It now points to the *git command reference* for Visual Studio 2017 and 2015. (I don't know where the VS 2019 version can be found :confused:).

If this link is still not right, we should close the issue in favor of a bigger rework.

ben:
Yeah, that looks better. Thanks!

HonkingGoose:
Cheers! :sparkles:

PR #1457, opened 2020-07-02
Replace 'option' by more specific terms where needed for 'git log --pretty'

cbachhuber:
Another good finding by @mo-gul: currently, everything related to `--pretty` is called option. This is not ideal for distinguishing the option from its values and the format specifiers. To resolve, I suggest the following naming scheme:

- Option: e.g. `--pretty`, `--stat`
- Option value: e.g. `short`, `full` after `--pretty=`
- [Format specifier](http://www.cplusplus.com/reference/cstdio/printf/): e.g. `%H`, `%T` in `--pretty=format:""`

I've implemented my suggestion for the section related to `--pretty`. I didn't check the entire book; we might want to align other places for consistency.

### DIFF
--- a/book/02-git-basics/sections/viewing-history.asc
+++ b/book/02-git-basics/sections/viewing-history.asc
@@ -132,9 +132,9 @@ It also puts a summary of the information at the end.
 
 Another really useful option is `--pretty`.
 This option changes the log output to formats other than the default.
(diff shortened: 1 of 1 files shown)


ben:
Nice. It's more exact, and doesn't sound any more stilted. Thanks!

PR #1459, opened 2020-07-04
Change references of 'pu' to 'seen'

### LINKED ISSUE #1447: 'pu' has been renamed to 'seen'
The v2 version of the book accessible via the website.

**Describe the bug:**
The Git maintainer has renamed the `pu`branch to `seen`. See: https://public-inbox.org/git/xmqqimfid2l1.fsf@gitster.c.googlers.com/

The `pu` branch is referenced in the book in a few places. So, they need to updated.

sivaraam:
Fixes #1447 

Ok. So, I've updated references of 'pu' to 'seen'. I've mentioned that it was previously called 'pu' in one place as it has served the Git community for 15 years. So, I think it deserves a mention for some more time.

There are still 2 references to 'pu' left:
1. The following reference in the "Long-running branches" section of Chapter 3 "Git Branching".
    > Some larger projects also have a `proposed` or `pu` (proposed updates) branch that has integrated branches that may not be ready to go into the `next` or `master` branch.

    I'm not particularly convinced about changing that as its not a direct reference to the Git project's branch. 
2. In 'images/merging-workflows-2.png'. This one is a direct reference to the Git project's branch and should be changed. I'm not sure how to go about doing that, though.

I went ahead and rename the reference in the, yet to be used, SVG image with hopes that it would be useful. I see that there's #1288 about using SVG instead of the PNG images. If that would be merged soon, I think the changes in this PR are enough :) If not, let me know how the PNG image should be updated.

### DIFF
--- a/book/05-distributed-git/sections/maintaining.asc
+++ b/book/05-distributed-git/sections/maintaining.asc
@@ -343,7 +343,7 @@ Then, when the codebase on that branch is stable and passes tests, you merge it 
 ===== Large-Merging Workflows
 
 (((workflows, "merging (large)")))
-The Git project has four long-running branches: `master`, `next`, and `pu` (proposed updates) for new work, and `maint` for maintenance backports.
+The Git project has four long-running branches: `master`, `next`, and `seen` (formerly 'pu' -- proposed updates) for new work, and `maint` for maintenance backports.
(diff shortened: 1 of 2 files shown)


ben:
The SVG switchover requires some work on the git-scm.com side to make sure the transition is smooth, but you've done the right thing here. Thanks!

PR #1472, opened 2020-07-30
Make commit name consistent

DustInCompetent:
The example commit `f7f3f6d Change my name a bit` had a different message in one of the sections


ben:
Nice catch!

Issue #1462, opened 2020-07-15
Discussion point: master terminology


paulclarkin:
I've had a look but can't see this already discussed - apologies if it has.

With Microsoft/GitHub recently discussing the removal of terms such as "master/slave"/"whitelist/blacklist" from their systems, including Github, could or should the term be review and/or removed from the Git Book and replaced with a term such as "main"?

I appreciate this would be a large task so might have to be done in stages.

ben:
I'm absolutely on board with this kind of effort, and I'd love to take steps towards removing this terminology. There are parts of this which are under our control, and parts that are not. Until Git changes the default branch name (or lets you customize it for `git init`), the word "master" will need to continue to appear in the text.

Here are the things we can do in the meantime:

- Add a note that describes how to set the default branch name
  - ...using a template repo for `git init`
  - ...using `git branch -m` for existing repos
- Explicitly change the branch name when doing `git init` in examples
- Assume the remote branch name is not "master" for examples with cloned repos

We do need to be careful to protect the newbie experience though. I don't want someone to get frustrated because they ran `git init` and the example text assumes that the fresh default branch is called `main` with no explanation.

HonkingGoose:
It turns out that changing from `master` to `main` is not as hard as it appears:

1. `git branch -m master main`
2. `git push -u origin main`
3. Then change the default branch in GitHub settings for the repository.
4. Change settings for continuous integration (Travis, GitHub Actions, Azure Pipeline).

So that takes care of the problem if you're only using the repository as a backup.

If you collaborate with others, the people with a fork/clone will need to make some changes too:

```
git checkout master
git branch -m master main
git fetch
git branch --unset-upstream
git branch -u origin/main
git symbolic-ref refs/remotes/origin/HEAD refs/remotes/origin/main
```

---

Yeah, running `git init` from your local terminal will still give a `master` main branch.

I don't know when GitHub themselves will use the `main` branch if you create a new repository from within their interface. 

The starter GitHub Actions have **just** been rejigged to use a "$default_branch" token instead of a hard-coded `master` branch.

---

Sources:

https://www.hanselman.com/blog/EasilyRenameYourGitDefaultBranchFromMasterToMain.aspx
https://github.com/actions/starter-workflows/pull/590

PR #1476, opened 2020-08-08
Create section: branch renaming

HonkingGoose:
Add a new section dealing with renaming branches containing:

- Heading: "Changing a branch name"
- Subheading: "Changing the master branch name".
- The command to change the branch name.
- A big [WARNING] message, so that users are not going to change their `master` branch willy-nilly.
- Give advice on the things that will need to be updated after changing the default branch.

This is a rough draft, and will need some feedback to get it ready for merging.
I also think some other files will need to be changed as well, so that it's all properly referencing each other. I'm really not sure where those references should be. So I could use some help with that. 😄 

This PR will help with issue: #1462


book/03-git-branching/sections/branch-management.asc
```diff
@@ -73,3 +73,42 @@ $ git branch --no-merged master
...
+
+==== Changing a branch name
+
+You can change a branch name with the `git branch -m` command.
+If you want to collaborate with others, on the newly changed branch, you'll need to push it.
+
+[source,console]
+----
+$ git branch -m bad-branch-name corrected-branch-name
+$ git push -u origin corrected-branch-name
```
ben, line 85: I think the end result here would be that both `bad-branch-name` and `corrected-branch-name` exist on the remote? `git branch -m` only renames the branch locally, so we may want to get a bit more in depth there — the concept that my branches and the remote's branches can be different that trips up a lot of newcomers.

book/03-git-branching/sections/branch-management.asc
```diff
@@ -73,3 +73,42 @@ $ git branch --no-merged master
...
+After changing the name, you'll need to push the renamed branch so that others can see it too.
+Other users with a clone/fork of your repository will need to change their default branch as well.
+
+[source,console]
+----
+$ git branch -m master main
+$ git push -u origin main
+----
+
+Things to update after pushing the new default branch:
```
ben, line 108: This feels a tiny bit short. Maybe a little more?

```suggestion
That changes the branch name in your repo, and makes the new name available in the remote.
Now you have a few more tasks in front of you to complete the transition:
```

book/03-git-branching/sections/branch-management.asc
```diff
@@ -73,3 +73,42 @@ $ git branch --no-merged master
   featureB
 ----
 ====
+
+==== Changing a branch name
+
+You can change a branch name with the `git branch -m` command.
+If you want to collaborate with others, on the newly changed branch, you'll need to push it.
```
ben, line 80: Ooh, one last thing: you've got one too many commas in here.

```suggestion
If you want to collaborate with others on the newly changed branch, you'll need to push it.
```

HonkingGoose: Whoops, you're right. It's still a bit hard to figure out where a comma should go, and where it should be omitted... 😕 

I usually err on the side of having too many, than having too few...

book/03-git-branching/sections/branch-management.asc
HonkingGoose: Yes, that's better than what I made. In the commit it goes! 😄

book/03-git-branching/sections/branch-management.asc
HonkingGoose: Agree, I'll go expand on this some more, and try a move on a new test repo to find out how the remote and local branch interact. That way I can document what happens.

HonkingGoose: >I think the end result here would be that both bad-branch-name and corrected-branch-name exist on the remote? 

Yes that is correct, I've just tested it.
The local bad branch is renamed, then you push the local correct branch up.
The remote now has both branches, both the correct one and the bad one.
The user then needs to `push origin --delete bad-branch-name` to get rid of the bad branch.


> so we may want to get a bit more in depth there — the concept that my branches and the remote's branches can be different that trips up a lot of newcomers.

I've added a overview of the branches just after pushing the correct branch, I hope this helps explain what just happened.

I'm marking this as resolved, because the review is now out of date. We'll continue this conversation on the newer version of the text 😉 .

HonkingGoose:
Hi @ben,

I noticed I merged in other work by accident on my push with newer stuff, I'll go fix that now...

Man writing technical docs is hard... especially for something as complicated as changing the _master_ branch.
I did some more work on the text, hopefully this is better than what we had just before. I still have some questions though.

Should we include links to GitHub documentation, for things like:

- [Changing the default branch](https://docs.github.com/en/github/administering-a-repository/setting-the-default-branch)
- [Protected branches ](https://docs.github.com/en/github/administering-a-repository/about-protected-branches#branch-protection-settings)

Should we mention things like _Dependabot/Travis CI_ by name, to help users understand what kind of things might break?

Do we want a vendor neutral text, or do we want to mention GitHub specific gotcha's?

HonkingGoose

book/03-git-branching/sections/branch-management.asc
```diff
@@ -73,3 +73,112 @@ $ git branch --no-merged master
...
+Now the bad branch name is fully replaced with the corrected branch name.
+
+===== Changing the master branch name
+
+[WARNING]
+====
+Changing the name of a branch like master/main/mainline/default will break the integrations, services, helper utilities and build/release scripts that your repository uses.
+Read the entire section before making any changes.
+Do not attempt this move without first consulting with your co-workers.
+Check beforehand where the old branch name is used in your project by doing a thorough search.
```
ben, line 135: A small flow improvement (I hope):

```suggestion
Before you do this, make sure you consult with your collaborators.
Also make sure you do a thorough search through your repo and update any references to the old branch name in your code or scripts.
```

book/03-git-branching/sections/branch-management.asc
```diff
@@ -73,3 +73,112 @@ $ git branch --no-merged master
...
+Other collaborators will continue to use the _master_ branch as the base of their work, until you make some further changes.
+
+Now you have a few more tasks in front of you to complete the transition:
+* Project dependency update helper configuration files.
+* Test runner configuration files.
+* Build scripts.
+* Release scripts.
+* Repository configuration settings (like the default branch that is targeted for pull-requests, branch-protection measures, etc).
+* References to the old branch in documentation.
+* Close/merge pull requests that target the old branch
```
ben, line 178: Just a few tweaks:

```suggestion
* Any projects that depend on this one will need to update their code and/or configuration.
* Update any test-runner configuration files.
* Adjust build and release scripts.
* Redirect settings on your repo host for things like thee repo's default branch, merge rules, and other things that match branch names.
* Update references to the old branch in documentation.
* Close or merge any pull requests that target the old branch.
```

HonkingGoose: I wasn't sure on how verbose to be. Your text is clearer, so let's use your suggestions!

HonkingGoose:
Ready for review and possibly a merge.

PR #1484, opened 2020-08-09
Create section: drop commit with git rebase -i

### LINKED ISSUE #824: Chapter 7, section on rebasing, should add info on "drop" command
Apparently, in addition to all of the other interactive rebase commands, we now have "drop". I'll let you handle that.

HonkingGoose:
Issue #824 says that we should add a section on dropping commits with `git rebase -i` script workflow.
This PR adds this section.

While writing the section, I noticed how easy it is to get to a broken state by dropping a commit:
- Do not drop the first commit in the list, because that breaks the chain of references for the following commits (what should commit 2 point to if commit 1 is gone???)
- Do not drop a commit that has work that other commits depend on.

And what if a reader succeeds in dropping a commit, on purpose/accidentally, and then panicking that their work is gone forever. It can be found in the reflog of-course, but panicked readers will not know or have the ability at that point to use the reflog to retrieve their dropped commit.

If the rebase with the dropped commit is not successful, the reader will enter into a inconsistent state, and either need to manually fix the problem, or back out of the rebase with `git rebase --abort`.

Given the above problems, I'm not sure whether the book should point users to such a "dangerous" thing as dropping a commit at all.... I'll let the maintainers decide on whether this belongs in the book or not.

Merging this fixes #824.


book/07-git-tools/sections/rewriting-history.asc
```diff
@@ -299,6 +299,34 @@ This changes the SHA-1s of the three most recent commits in your list, so make s
 Notice that the last commit (`f7f3f6d`) in the list is unchanged.
 Despite this commit being shown in the script, because it was marked as ``pick'' and was applied prior to any rebase changes, Git leaves the commit unmodified.
 
+==== Deleting a commit
+
+If you want to get rid of a commit, you can delete it using the `rebase -i` script.
+In the list of commits, put the word ``drop'' before the commit you want to delete:
```
ben, line 305:
```suggestion
If you want to get rid of a commit, you can delete it using the `rebase -i` script.
In the list of commits, put the word ``drop'' before the commit you want to delete (or just delete that line from the rebase script):
```

book/07-git-tools/sections/rewriting-history.asc
```diff
@@ -299,6 +299,34 @@ This changes the SHA-1s of the three most recent commits in your list, so make s
...
+
+Avoid doing this:
+[source,console]
+----
+drop 5aecc10 This commit is broken
+pick c6be4c9 This commit is OK but depends on changes made in 5aecc10
+----
+
+Git will complain about the broken state, and ask you to resolve the conflicts manually.
+You can also just back out of the rebase by using `git rebase --abort`.
```
ben, line 328: Here's an alternate version, let me know what you think. It avoids the specific advice about removing the first commit in a stream (there are situations where that will work just fine) in favor of more general advice.

```suggestion
Because of the way Git builds commit objects, deleting or altering a commit will cause the rewriting of all the commits that follow it.
The further back in your repo's history you go, the more commits will need to be recreated.
This can cause lots of merge conflicts if you have many commits later in the sequence that depend on the one you just deleted.

If you get partway through a rebase like this and decide it's not a good idea, you can always stop.
Type `git rebase --abort`, and your repo will be returned to the state it was in before you started the rebase.

If you finish a rebase and decide it's not what you want, you can use `git reflog` to recover an earlier version of your branch.
See <<ch10-git-internals#_data_recovery>> for more information on the `reflog` command.
```

HonkingGoose: That's way better than what I had. 😄 In it goes!

I did not know that there are situations in which removing the first commit in a stream *won't result* in breakage.
I'm not sure where that explanation can be found? Is this covered in the book already somewhere? It would sure be nice to have that explanation somewhere, if it's not already in... 😄

ben: If you have three commits, and commit `A` only touches file `xyz`, and commits `B` and `C` don't touch that file at all, then you won't hit conflicts by removing `A` from the stream.

Git doesn't have "special" commits except maybe a commit without parents. _That_ one isn't accessible from a `rebase -i` script, but you can remove it by rebasing all the subsequent commits onto another parentless commit. **This** is an obscure corner-case though, so I'd hesitate to write about it in the book.

PR #1486, opened 2020-08-10
Edit Tim Pope's commit guidelines for clarity

### LINKED ISSUE #696: Adding a footnote regarding the a statement in the commit template
In the commit template found in the _Commit Guidelines_ subsection of the _Contributing to a Project_ section, there's a line that reads like this,

**The blank line separating the summary from the body is critical (unless you omit the body entirely); tools like rebase can get confused if you run the two together.**

It would be better if the reason, why tools like rebase get confused was stated probably as a footnote (concisely, of course). It would help the reader to be more aware of it.

HonkingGoose:

## Changes

- Clarify that it's the end user that will get confused with rebasing when the body and summary of the commit message are mushed together.
- Edit the line that links to the original Tim Pope blog post, stating that we adapted it. We're no longer fully quoting him.

## Context

Fixes #696

@ben I'm not sure if editing the source text that Tim Pope created is the right thing to do here?  Seeing as we quoting his words, and then edit them later... Should we use a footnote/callout instead? Or just leave it be and just close both this PR and the linked issue?

I noticed there might be two options to do this more cleanly, a callout `<1>` or a `footnote:[]`. I do not know if those are supported within a `[source, text]` block?

Source: Asciidoctor manual for [callouts](https://asciidoctor.org/docs/user-manual/#callouts) and for [footnotes](https://asciidoctor.org/docs/user-manual/#user-footnotes).


book/05-distributed-git/sections/contributing.asc
```diff
@@ -60,7 +60,7 @@ Getting in the habit of creating quality commit messages makes using and collabo
 As a general rule, your messages should start with a single line that's no more than about 50 characters and that describes the changeset concisely, followed by a blank line, followed by a more detailed explanation.
 The Git project requires that the more detailed explanation include your motivation for the change and contrast its implementation with previous behavior -- this is a good guideline to follow.
 Write your commit message in the imperative: "Fix bug" and not "Fixed bug" or "Fixes bug."
-Here is a https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html[template originally written by Tim Pope]:
+Here is a https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html[template originally written by Tim Pope], and lightly adapted by us:
```
ben, line 63: One small flow improvement:

```suggestion
Here is a template you can follow, which we've lightly adapted from one https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html[originally written by Tim Pope]:
```

HonkingGoose: Yeah that's better, in it goes! 😄

HonkingGoose:
I think this is ready for the final review, and merge afterwards. 😄

PR #1489, opened 2020-08-16
Add tip section with link to git-send-email.io

HonkingGoose:

## Changes

- Add tip with a link to https://git-send-email.io/ in the _Contribution to the Public Project over Email_ part of the book

## Context

This website gives help on configuring your system (what packages to install on your Linux distribution), tells you how to configure your email client, and has a "sandbox repo" that you can target for your very first Git email patch. There are also extra tips and tricks.

I think this is a good match for what this section of the book is about. First you explain the high level overview of how things work, and give some examples in the book. And then the reader can read the tip, go to the website and give the whole shebang a try.

I think this section of the book makes the most sense. I you disagree, feel free to tell me where you want this tip, and I'll move it.

### DIFF
--- a/book/05-distributed-git/sections/contributing.asc
+++ b/book/05-distributed-git/sections/contributing.asc
@@ -785,6 +785,11 @@ References: <y>
 Result: OK
 ----
 
+[TIP]
+====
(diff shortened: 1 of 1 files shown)


ben:
I actually think this is the perfect place for it.

PR #1490, opened 2020-08-17
Fix unordered list to render properly

petrkajzar:
An unordered list without a leading blank line renders only as an inline text. Adding a new blank line is enough to make it render properly as an unordered list, as intended.

## Changes

- Add a new blank line in front of the unordered list to make it render properly.

## Context

- None.


HonkingGoose:
@petrkajzar Good catch! You're right, it's not rendering properly on the live site...

From the [Asciidoctor Writer's guide on lists](https://asciidoctor.org/docs/asciidoc-writers-guide/#lists-of-things):
> Blank lines are required before and after a list.

PR #1502, opened 2020-08-28
Add v before version number for consistency

### LINKED ISSUE #852: fix tag name in 2.6
https://git-scm.com/book/en/v2/Git-Basics-Tagging

The final example begins with "git checkout 2.0.0" but at the very bottom the code is assuming that the tag is named v2.0.0. I suppose they don't have to be consistent but it makes the most sense if they are, so it seems best to add a v to the 2.0.0 at the top.

HonkingGoose:

## Changes

- Add v before version number for consistency

## Context

Throughout the tagging section we prepend any version numbers with "v", this commit ensures that we're consistent throughout. This also closes issue #852 that found some inconsistencies with the way we use tags.

Fixes #852

### DIFF
--- a/book/02-git-basics/sections/tagging.asc
+++ b/book/02-git-basics/sections/tagging.asc
@@ -276,7 +276,7 @@ do so (now or later) by using -b with the checkout command again. Example:
 
 HEAD is now at 99ada87... Merge pull request #89 from schacon/appendix-final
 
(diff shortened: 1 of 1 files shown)


HonkingGoose:
Force pushed the current fixed mainline branch into this branch so that we can see if the build is working or not.

Issue #1109, opened 2018-09-18
Updating outdated screenshots and installation methods for GitLab


borivojetasovac:
I noticed a few outdated GitLab screenshots in the 4.8 chapter ("Git on the Server - GitLab") of the book. Also, I'd like to update the installation guidelines to match the current best practices.

Any feedback/suggestions before I start?

ben:
This would be great! I know GitLab has been shipping like crazy, so it's no surprise that section is outdated.

HonkingGoose:
I've updated the text covering GitLab with #1504, but we still need newer GitLab pictures. I don't run a GitLab server, so I cannot take the pictures we need.

Can somebody take pictures and add them in a new pull-request that closes this issue? Thanks! :+1:

PR #1504, opened 2020-08-28
Update GitLab section text

HonkingGoose:

## Changes

- Update link to GitLab CE readme as the repository linked to has been moved
- Update GitLab installation methods and recommendations to be inline with current GitLab methods and recommendations
- Some textual fixes for clarity
- Use the active voice instead of the passive voice
- Remove third party Bitnami section and image as I think we should only provide official resources

## Context

Helps with #1109 as it updates the installation methods and guidance a bit, but this pull does not include newer pictures for the GitLab interface (I don't run a GitLab server...).

The installation methods update is based on the [GitLab CE official docs](https://docs.gitlab.com/ce/install/README.html) and on the [GitLab self-managed installation page](https://about.gitlab.com/install/).

I've tried to keep each change limited to a single commit, so that we can revert changes you don't want easily.


tfiers:
(Install step in CI fails, seems unrelated to this PR)

HonkingGoose:
Force pushed the current fixed mainline branch into this branch so that we can see if the build is working or not.

HonkingGoose:
I'm quite happy with this now, it's ready for a review. :smile: :+1:

Issue #1496, opened 2020-08-25
Breaking builds: Kindlegen dependency no longer available


HonkingGoose:
**Which version of the book is affected?**

:warning: This issue affects all builds! :warning: 
Problem will be present on all future builds/releases once the Travis CI cache is expired. Problem is with source files.

**Describe the bug:**

The 'kindlegen' binary that we depend on to build/validate the Kindle files is no longer available to download from the rubygems.org website. Because this dependency is no longer available, Travis CI cannot build new releases of the book.

I'm specifically talking about this entry in the Gemfile:
https://github.com/progit/progit2/blob/8b361da3b1d4424180d5ba3a204f8616436bdd01/Gemfile#L16

**Steps to reproduce:**

1. Try to run a build on Travis CI.
2. Build fails with:

```
mAn error occurred while installing kindlegen (3.0.5), and Bundler

cannot continue.

Make sure that `gem install kindlegen -v '3.0.5' --source

'https://rubygems.org/'` succeeds before bundling.

In Gemfile:

  kindlegen

The command "bundle install" failed and exited with 5 during .
```

Example of failed build: https://travis-ci.org/github/HonkingGoose/progit2/builds/721079601

**Expected behavior:**

Build passes.

**Screenshots:**

**Additional context:**

Upstream issue reports: https://github.com/tdtds/kindlegen/issues/42 https://github.com/asciidoctor/asciidoctor-epub3/issues/363
We also have some other issues outstanding with dependencies at: #1373 #1358 

Amazon says to use Kindle Previewer instead, to get the same functionality: 
https://www.amazon.com/gp/feature.html?docId=1000765211
This will not work for our project, as according to @slonopotamus the Kindle Previewer does not exist for Linux.

This problem is also present when I try to download the Kindlegen dependency locally. So it's not just Travis CI that's broken...

slonopotamus:
I've also created asciidoctor/asciidoctor-epub3#363 because chances that there will be any fix to tdtds/kindlegen#42 are very low. `kindlegen` gem is just a thin wrapper that downloads KindleGen binary to comply with EULA, but now there isn't anything to download unless someone publishes KindleGen binaries on their own website.

And the problem with Kindle Previewer is that it doesn't exist for Linux.

HonkingGoose:
Thanks for your feedback @slonopotamus! :smile: 
I've edited my top post to include your findings.

[...]

PR #1506, opened 2020-08-28
Link to more precise "What is Git" page in section 3.1 ("Branches in a Nutshell")

tfiers:
[Section 3.1, "Branches in a Nutshell"](https://git-scm.com/book/en/v2/Git-Branching-Branches-in-a-Nutshell) talks about snapshots and hashes, and refers to [this page](https://git-scm.com/book/en/v2/Getting-Started-About-Version-Control) for more info on them. These topics don't appear on that page however. The author is probably referring to [section 1.3, "What is Git"](https://git-scm.com/book/en/v2/Getting-Started-What-is-Git%3F). This page is from the same chapter as is currently linked, but I think a more precise link would help the casual reader more.

This PR changes the link to point to section 1.3, "What is Git", instead of to the entire chapter, which takes the reader to section 1.1, "About Version Control".

## Changes

- Add ID (`[[_what_is_git]]`) to section 1.3
- Link to this ID in section 3.1

### DIFF
--- a/book/01-introduction/sections/what-is-git.asc
+++ b/book/01-introduction/sections/what-is-git.asc
@@ -1,3 +1,4 @@
+[[_what_is_git]]
 === What is Git?
 
 So, what is Git in a nutshell?
--- a/book/03-git-branching/sections/nutshell.asc
+++ b/book/03-git-branching/sections/nutshell.asc
@@ -3,13 +3,13 @@
 
 To really understand the way Git does branching, we need to take a step back and examine how Git stores its data.
 
-As you may remember from <<ch01-getting-started#ch01-getting-started>>, Git doesn't store data as a series of changesets or differences, but instead as a series of _snapshots_.
+As you may remember from <<ch01-getting-started#_what_is_git>>, Git doesn't store data as a series of changesets or differences, but instead as a series of _snapshots_.
 
 When you make a commit, Git stores a commit object that contains a pointer to the snapshot of the content you staged.
 This object also contains the author's name and email address, the message that you typed, and pointers to the commit or commits that directly came before this commit (its parent or parents): zero parents for the initial commit, one parent for a normal commit, and multiple parents for a commit that results from a merge of two or more branches.
 
 To visualize this, let's assume that you have a directory containing three files, and you stage them all and commit.
-Staging the files computes a checksum for each one (the SHA-1 hash we mentioned in <<ch01-getting-started#ch01-getting-started>>), stores that version of the file in the Git repository (Git refers to them as _blobs_), and adds that checksum to the staging area:
+Staging the files computes a checksum for each one (the SHA-1 hash we mentioned in <<ch01-getting-started#_what_is_git>>), stores that version of the file in the Git repository (Git refers to them as _blobs_), and adds that checksum to the staging area:
 
 [source,console]
 ----


tfiers:
(Install step in CI fails, seems unrelated to this PR)

HonkingGoose:
>(Install step in CI fails, seems unrelated to this PR)

That's right, you're not breaking something here! :+1: 

In a nutshell: The Travis CI is broken because the `kindlegen` dependency cannot be installed anymore. See #1496 and #1497 for more info.

tfiers:
Huh, these internal links seem broken in the .html and .pdf of the [release of this PR](https://github.com/progit/progit2/releases/tag/2.1.254):

<div align="center">
</div>
<br>

Not sure why, as I've used the same syntax as [here](https://github.com/progit/progit2/blob/master/book/02-git-basics/sections/getting-a-repository.asc) and [here](https://github.com/progit/progit2/blob/master/book/04-git-server/sections/git-on-a-server.asc), eg.

tfiers:
OK, found the problem: https://github.com/asciidoctor/asciidoctor-epub3/issues/139

Willl look into it

HonkingGoose:
I've tried comparing what you did with other known working examples by doing a plain-text comparison and trying to think of issues. But cannot actually figure it out, because I cannot get a proper build tool-chain working on my end... :disappointed:

I don't know what causes this issue. So you'll have to wait until @ben or @jnavila can chime in on why this happens.

tfiers:
Yeah, can't get toolchain to work locally either (on Windows). Might continue trying on Windows Subsystem for Linux.

One difference with my patch versus the other pages is that, coincidentally, in my edited paragraph there are subsequent `_` marks for emphasis. This is not the case after the `#_getting_git_on_a_server` link [here](https://github.com/progit/progit2/edit/master/book/02-git-basics/sections/getting-a-repository.asc) eg.

HonkingGoose:
That does seem to be a new error that was not there before though.
And if other sections are also broken I would expect more errors like that showing up.

tfiers:
At a first look there do not seem to be any other paragraphs in the book where a `_` emphasis mark follows an internal link, within the same paragraph. So seems like we discovered a rare bug here.

jnavila:
You shouldn't use anchor names starting with "_", because these are reserved for automatically generated links. The automatically generated links are made from the text of the title in lower case with spaces converted to underscores, prefixed with an underscore, which happens to be exactly the chosen anchor name. So, I guess, there's a clash in anchor definitions, which leads to no anchor created. Change the anchor name for `ch_wha_is_git`, and it will work.

HonkingGoose:
Thanks for chiming in @jnavila! And I learned something new today! :smile:

tfiers:
For future reference: after testing, it seems like it doesn't matter what the anchor is called (whether it matched the auto ID or not): as long as it starts with an underscore and there's some `_emphasis_` in the same paragraph following a reference to that anchor, we'll get the error.

HonkingGoose:
What a weird error. :confused: 

 Let's hope we can get rid of this after bumping us to Asciidoctor 2. And that our improved build system that I'm working on can help catch such error before deploying them to `master`. But that's for the future, so I'm making no claims of success on neither of those. :smile:

Issue #1441, opened 2020-06-22
[Suggestion]: Update 2.4 Git Basics - Undoing Things section with `git restore`


prosoitos:
I'd be happy to submit a PR on this if people are ok with the change.

[2.4 Git Basics - Undoing Things](https://git-scm.com/book/en/v2/Git-Basics-Undoing-Things) has fallen out of date now that `git restore <file>` is suggested instead of `git checkout -- <file>` to get rid of changes in unstaged files.

The paragraph "Unmodifying a Modified File" of that page, which predates the launch of the new command `git restore`, does not reflect the current version of Git anymore. I would be happy to update this, but I wanted to make sure the book authors agreed with this first (particularly since the old command still works and there is an experimental warning in the `git-restore` man page).

prosoitos:
Actually, I just realized that the section "Unstaging a Staged File" would require a similar update since the recommended way to unstage a file, since the addition of the `git restore` command, is `git restore --staged <file>` instead of the old `git reset HEAD <file>`.

ben:
Yeah, that sounds like a positive change for sure. I'd love to see a PR!

prosoitos:
Excellent :slightly_smiling_face: 

I'll do this in the coming week or so.

HonkingGoose:
Hi @prosoitos good catch!

I noticed the change from `git reset` and `git checkout --` to `git restore` would touch quite a lot of files:

- String `git reset` gets 52 hits in 16 `.asc` files.
- String `git checkout --` 24 hits in 9 `.asc` files.

I also think that the old undo actions have more references that won't show up with only those 2 string searches. So I strongly suspect that you would need to search for all hits on `git checkout` and `git reset` to ensure you got 'em all... :smile: 

If you want, we could split the work between us, and work together on a branch? So that you're not having to do this bunch of work alone?

prosoitos:
Thanks for the heads-up @HonkingGoose.

On my side, I also realized that `git switch`, which is the other command `git checkout` got split into in 2.4 isn't included in the book either. It would make sense to add both of them at the same time.

[...]

PR #1511, opened 2020-08-29
Create section git restore

### LINKED ISSUE #1487: Inconsistency: Git restore is only used once, rest of book uses reset HEAD

Problem is with the source files, present on current `master`.

**Describe the bug:**

Inconsistency in the way to unstage a staged file.
Our examples use use `git reset HEAD <file to unstage>`.
The status output we show is using `git reset HEAD`, as that is Git's prior advice before the `git restore` command became a thing.

There's one place where the book uses `git restore` and it's introduced with commit: https://github.com/progit/progit2/commit/dc15a308c5ce870699706849a5975dea178cce36

**Steps to reproduce:**

1. Search for hits on the string `git reset HEAD` in the book, notice it's used everywhere for `git status` and unstaging examples the book uses.
2. Search for hits on the string `git restore`, notice it's only used once.

**Expected behavior:**

I expect the book to present a single way to unstage, and be consistent with it, so that readers are never confused.

**Options to fix this:**
1. Revert the merge that resulted in the inconsistency: https://github.com/progit/progit2/pull/1338
2. Revert the merge that resulted in the inconsistency: https://github.com/progit/progit2/pull/1338 plus create a TIP block in an appropriate place in the book with the hint that `git restore` can be used instead of `git reset HEAD`.
3. Change over the entire book to use the newer `git restore` command, and put a note up in a appropriate place with something to the effect of: "If `git restore` doesn't work for you, try `git reset HEAD`.

I'll wait for the maintainers feedback before making a pull request to fix it in one of the three ways listed.

**Context:**

This kind of issue also crops up with: https://github.com/progit/progit2/issues/1441

HonkingGoose:

## Changes

- Create section on `git restore`

## Context

Fixes #1487 as we now have a section explaining what `git restore` is.
Read #1487 for the discussion we had leading up to this pull request.

This issue is also related:  #1441 but I'm not closing it with this pull, because that issue also talks about the `git switch` command, which this pull request does not add.

This is a rough draft, that basically copies the two undo sections that used `git reset` and replaced the examples with `git restore`.

I'm of course open to your feedback to improve the text. You always have great suggestions! :smile: :+1: 

---

## Some questions/thoughts:

Should we perhaps change the example a bit more, so that we have less of a deja-vu experience? Or would that confuse readers even more?

I feel what I have now is quite heavy, and duplicates a lot of stuff.

And maybe rewrite this sentence, I really don't like this, it's a bit too complicated to understand at first read:
> The nice part is that the command you use to determine the state of those two areas also reminds you how to undo changes to them.

It's a lot of words to basically say: `git status` is your friend and gives you helpful advice on how to undo stuff. :smile_cat:

### DIFF
--- a/book/02-git-basics/sections/undoing.asc
+++ b/book/02-git-basics/sections/undoing.asc
@@ -146,3 +146,83 @@ If you would like to keep the changes you've made to that file but still need to
 Remember, anything that is _committed_ in Git can almost always be recovered.
 Even commits that were on branches that were deleted or commits that were overwritten with an `--amend` commit can be recovered (see <<ch10-git-internals#_data_recovery>> for data recovery).
 However, anything you lose that was never committed is likely never to be seen again.
+
+[[undoing_git_restore]]
+==== Undoing things with git restore
+
+Git version 2.25.0 introduced a new command: `git restore`.
+It's basically a alternative to `git reset` which we just covered.
+From Git version 2.25.0 onwards, Git will use `git restore` instead of `git reset` for many undo operations.
+
+Let's retrace our steps, and undo things with `git restore` instead of `git reset`.
+
+===== Unstaging a Staged File with git restore
+
+The next two sections demonstrate how to work with your staging area and working directory changes with `git restore`.
+The nice part is that the command you use to determine the state of those two areas also reminds you how to undo changes to them.
+For example, let's say you've changed two files and want to commit them as two separate changes, but you accidentally type `git add *` and stage them both.
+How can you unstage one of the two?
+The `git status` command reminds you:
+
+[source,console]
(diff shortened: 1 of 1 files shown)


HonkingGoose:
Hi @ben, this is ready for review. It's still a draft, and too rough to merge yet. I'll mark it ready for review after fixing it with your feedback. :smile: :wink:

ben:
Yeah, this looks great!

HonkingGoose:
Changes:

- Squash already approved changes into one commit and rebase on top of current `master`.
- Add warning about `git restore --staged <file>` throwing away your local work.

ben:
:sparkles:

PR #1539, opened 2020-10-05
removed `--add` from TextEdit core.editor config

### LINKED ISSUE #1537: Appendix C: core.editor command for TextEdit incorrectly uses --add

The online book: https://git-scm.com/book/en/v2/Appendix-C:-Git-Commands-Setup-and-Config

**Describe the bug:**

The command to set your core.editor setting for Text Edit is listed as
`git config --global --add core.editor "open -W -n"`
it should be
`git config --global core.editor "open -W -n"` (no --add parameter)

**Steps to reproduce:**

View Appendix C and look at the line for TextEdit (macOS)

**Expected behavior:**

The command should be `git config --global core.editor "open -W -n"`, which is in line with all the other core.editor commands and won't accidentally add (potentially) two entries for the setting.

**Desktop:**

- Operating system: Windows 10 Pro
- Browser/application: Chrome
- Browser/application version: 85.0.4183.121 (64 Bit)

Sophismata:

## Changes

- Removed the erroneous --add parameter from the example core.editor config used for TextEdit.

## Context

Fixes #1537

### DIFF
--- a/C-git-commands.asc
+++ b/C-git-commands.asc
@@ -62,7 +62,7 @@ Accompanying the configuration instructions in <<ch01-getting-started#_editor>>,
 |Scratch (Linux)|`git config --global core.editor "scratch-text-editor"`
 |Sublime Text (macOS) |`git config --global core.editor "/Applications/Sublime\ Text.app/Contents/SharedSupport/bin/subl --new-window --wait"`
 |Sublime Text (Windows 64-bit) |`git config --global core.editor "'C:\Program Files\Sublime Text 3\sublime_text.exe' -w"` (Also see note below)
-|TextEdit (macOS)|`git config --global core.editor "open -W -n"`
+|TextEdit (macOS)|`git config --global core.editor "open --wait-apps --new -e"`
 |Textmate |`git config --global core.editor "mate -w"`
 |Textpad (Windows 64-bit) |`git config --global core.editor "'C:\Program Files\TextPad 5\TextPad.exe' -m` (Also see note below)
 |UltraEdit (Windows 64-bit) | `git config --global core.editor Uedit32`
--- a/C-git-commands.asc
+++ b/C-git-commands.asc
@@ -62,7 +62,7 @@ Accompanying the configuration instructions in <<ch01-getting-started#_editor>>,
 |Scratch (Linux)|`git config --global core.editor "scratch-text-editor"`
 |Sublime Text (macOS) |`git config --global core.editor "/Applications/Sublime\ Text.app/Contents/SharedSupport/bin/subl --new-window --wait"`
 |Sublime Text (Windows 64-bit) |`git config --global core.editor "'C:\Program Files\Sublime Text 3\sublime_text.exe' -w"` (Also see note below)
-|TextEdit (macOS)|`git config --global --add core.editor "open -W -n"`
+|TextEdit (macOS)|`git config --global core.editor "open -W -n"`
 |Textmate |`git config --global core.editor "mate -w"`
 |Textpad (Windows 64-bit) |`git config --global core.editor "'C:\Program Files\TextPad 5\TextPad.exe' -m` (Also see note below)
 |UltraEdit (Windows 64-bit) | `git config --global core.editor Uedit32`


Sophismata:
Apologies; I wanted to make a quick PR but goofed it up. This is the correction for issue #1537, but when I attempted to fix the branch name of my first PR I mucked up the naming a second time.

HonkingGoose:
Thanks for contributing!

I always goof up changing the branch name as well! :smile: It's way too easy to miss changing the default `Username-patch-x` branch name, don't feel bad about that. :+1:

C-git-commands.asc
```diff
@@ -62,7 +62,7 @@ Accompanying the configuration instructions in <<ch01-getting-started#_editor>>,
 |Scratch (Linux)|`git config --global core.editor "scratch-text-editor"`
 |Sublime Text (macOS) |`git config --global core.editor "/Applications/Sublime\ Text.app/Contents/SharedSupport/bin/subl --new-window --wait"`
 |Sublime Text (Windows 64-bit) |`git config --global core.editor "'C:\Program Files\Sublime Text 3\sublime_text.exe' -w"` (Also see note below)
-|TextEdit (macOS)|`git config --global --add core.editor "open -W -n"`
+|TextEdit (macOS)|`git config --global core.editor "open -W -n"`
```
HonkingGoose, line 65:
```suggestion
|TextEdit (macOS)|`git config --global core.editor "open --wait --new-window"`
```
While we're changing this line anyway: we might want to take the opportunity to use more descriptive flags.
We're already using the long form flags for the other editors as well (but we're not fully consistent either...).

I'll let @ben the head maintainer decide on whether or not my suggestion is any good. :+1:

Sophismata: Thanks! I feel a bit better about that now 😁. Let me know if you need anything else from me, long form flags looks like a good improvement to me.

HonkingGoose: Can you confirm if these changes work on TextEdit? I'm not sure if they actually have long form flags for their editor....

Sophismata: I can't right now, I don't have a Mac set up at the moment. I've got a blank one sitting around that I need to set up and configure, I can do so this week if need be.

HonkingGoose: I'll go find some documentation on TextEdit and see if I can find some long form flags instead.

ben: Sorry for the delay. _gestures vaguely to the world_

I finally got around to trying this. Here's what `open` supports:

```
❯ open --help
open: unrecognized option `--new-window'
Usage: open [-e] [-t] [-f] [-W] [-R] [-n] [-g] [-h] [-s <partial SDK name>][-b <bundle identifier>] [-a <application>] [filenames] [--args arguments]
Help: Open opens files from a shell.
      By default, opens each file using the default application for that file.
      If the file is in the form of a URL, the file will be opened as a URL.
Options:
      -a                Opens with the specified application.
      -b                Opens with the specified application bundle identifier.
      -e                Opens with TextEdit.
      -t                Opens with default text editor.
      -f                Reads input from standard input and opens with TextEdit.
      -F  --fresh       Launches the app fresh, that is, without restoring windows. Saved persistent state is lost, excluding Untitled documents.
      -R, --reveal      Selects in the Finder instead of opening.
      -W, --wait-apps   Blocks until the used applications are closed (even if they were already running).
          --args        All remaining arguments are passed in argv to the application's main() function instead of opened.
      -n, --new         Open a new instance of the application even if one is already running.
      -j, --hide        Launches the app hidden.
      -g, --background  Does not bring the application to the foreground.
      -h, --header      Searches header file locations for headers matching the given filenames, and opens them.
      -s                For -h, the SDK to use; if supplied, only SDKs whose names contain the argument value are searched.
                        Otherwise the highest versioned SDK in each platform is used.
```

So this would work for a `core.editor` config, at least on Catalina:

```
open --wait-apps --new -e
```

HonkingGoose, line 65:
```suggestion
|TextEdit (macOS)|`git config --global core.editor "open --wait-apps --new -e"`
```

This suggestion implements @ben's `core.editor` config that he tried on Catalina.

Sophismata:
Thanks! Sorry for the late reply, I've been absolutely smashed at work and only just checked my notifications today. Do you need me to do anything further?

HonkingGoose:
Hi @Sophismata, thanks for adding the suggestion. I don't need you to do anything more. :wink: 

@ben I think this is ready for your review and merge.

ben:
:zap:

PR #1544, opened 2020-10-17
Fix typo in section 2.4

ardavast:


ben (approved):
Oof, good catch. Thanks!

PR #1545, opened 2020-10-21
Add note that SHA-1 actually has been broken

stuartpb:

### DIFF
--- a/book/07-git-tools/sections/revision-selection.asc
+++ b/book/07-git-tools/sections/revision-selection.asc
@@ -84,6 +84,8 @@ Here's an example to give you an idea of what it would take to get a SHA-1 colli
 If all 6.5 billion humans on Earth were programming, and every second, each one was producing code that was the equivalent of the entire Linux kernel history (6.5 million Git objects) and pushing it into one enormous Git repository, it would take roughly 2 years until that repository contained enough objects to have a 50% probability of a single SHA-1 object collision.
 Thus, a SHA-1 collision is less likely than every member of your programming team being attacked and killed by wolves in unrelated incidents on the same night.
 
+That being said, with hundreds and thousands of dollars' worth of computing power dedicated to the task, https://shattered.io/[two files with the same SHA-1 hash were produced in 2017]. Git has code https://github.com/git/git/blob/master/sha1dc/sha1.c[specifically to protect against this scenario].
+
 ====
 
 [[_branch_references]]
--- a/book/07-git-tools/sections/revision-selection.asc
+++ b/book/07-git-tools/sections/revision-selection.asc
@@ -82,10 +82,13 @@ That's 1,200 times the number of grains of sand on the earth.
 
 Here's an example to give you an idea of what it would take to get a SHA-1 collision.
 If all 6.5 billion humans on Earth were programming, and every second, each one was producing code that was the equivalent of the entire Linux kernel history (6.5 million Git objects) and pushing it into one enormous Git repository, it would take roughly 2 years until that repository contained enough objects to have a 50% probability of a single SHA-1 object collision.
-Thus, a SHA-1 collision is less likely than every member of your programming team being attacked and killed by wolves in unrelated incidents on the same night.
+Thus, an organic SHA-1 collision is less likely than every member of your programming team being attacked and killed by wolves in unrelated incidents on the same night.
 
-That being said, with hundreds and thousands of dollars' worth of computing power dedicated to the task, https://shattered.io/[two files with the same SHA-1 hash were produced in 2017]. Git has code https://github.com/git/git/blob/master/sha1dc/sha1.c[specifically to protect against this scenario].
-
+[NOTE]
+====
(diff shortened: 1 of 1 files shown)


book/07-git-tools/sections/revision-selection.asc
```diff
@@ -84,6 +84,8 @@ Here's an example to give you an idea of what it would take to get a SHA-1 colli
 If all 6.5 billion humans on Earth were programming, and every second, each one was producing code that was the equivalent of the entire Linux kernel history (6.5 million Git objects) and pushing it into one enormous Git repository, it would take roughly 2 years until that repository contained enough objects to have a 50% probability of a single SHA-1 object collision.
 Thus, a SHA-1 collision is less likely than every member of your programming team being attacked and killed by wolves in unrelated incidents on the same night.
 
+That being said, with hundreds and thousands of dollars' worth of computing power dedicated to the task, https://shattered.io/[two files with the same SHA-1 hash were produced in 2017]. Git has code https://github.com/git/git/blob/master/sha1dc/sha1.c[specifically to protect against this scenario].
+
```
ben, line 88: I like what you've got here, but let's make it flow a bit more. What do you think about this?

```suggestion
Thus, an organic SHA-1 collision is less likely than every member of your programming team being attacked and killed by wolves in unrelated incidents on the same night.

[NOTE]
====
There have been attempts to create a synthetic collision attack, including one documented at https://shattered.io/[].
Git is moving towards using SHA256 as the default hashing algorithm, which is much more resilient to collision attacks, and has code in place to help mitigate this attack (although it cannot completely eliminate it).
====
```

book/07-git-tools/sections/revision-selection.asc
```diff
@@ -82,8 +82,13 @@ That's 1,200 times the number of grains of sand on the earth.
...
 If all 6.5 billion humans on Earth were programming, and every second, each one was producing code that was the equivalent of the entire Linux kernel history (6.5 million Git objects) and pushing it into one enormous Git repository, it would take roughly 2 years until that repository contained enough objects to have a 50% probability of a single SHA-1 object collision.
-Thus, a SHA-1 collision is less likely than every member of your programming team being attacked and killed by wolves in unrelated incidents on the same night.
+Thus, an organic SHA-1 collision is less likely than every member of your programming team being attacked and killed by wolves in unrelated incidents on the same night.
 
+[NOTE]
+====
+There have been attempts to create a synthetic collision attack, including one documented at https://shattered.io/[].
+Git is moving towards using SHA256 as the default hashing algorithm, which is much more resilient to collision attacks, and has code in place to help mitigate this attack (although it cannot completely eliminate it).
+====
 ====
```
ben, line 92: Oof, just now realizing that we're already inside a note. Let's remove the nesting, I don't think it'll work:

```suggestion
There have been attempts to create a synthetic collision attack, including one documented at https://shattered.io/[].
Git is moving towards using SHA256 as the default hashing algorithm, which is much more resilient to collision attacks, and has code in place to help mitigate this attack (although it cannot completely eliminate it).
====
```

stuartpb:
I've adjusted the wording on that first line: "There have been attempts to create a synthetic collision attack" makes it sound like there've only been *attempts*, but no actual collisions. That was the case, when this book was written, but as of 2017, there have been *successful* attempts, at great expense, which is the point I'm trying to illustrate here: while it *is possible* to have two files with the same SHA-1 hash, it is something you have to *very* intentionally.

book/07-git-tools/sections/revision-selection.asc
```diff
@@ -82,8 +82,10 @@ That's 1,200 times the number of grains of sand on the earth.
 
 Here's an example to give you an idea of what it would take to get a SHA-1 collision.
 If all 6.5 billion humans on Earth were programming, and every second, each one was producing code that was the equivalent of the entire Linux kernel history (6.5 million Git objects) and pushing it into one enormous Git repository, it would take roughly 2 years until that repository contained enough objects to have a 50% probability of a single SHA-1 object collision.
-Thus, a SHA-1 collision is less likely than every member of your programming team being attacked and killed by wolves in unrelated incidents on the same night.
+Thus, an organic SHA-1 collision is less likely than every member of your programming team being attacked and killed by wolves in unrelated incidents on the same night.
 
+If you dedicate several thousands of dollars' worth of computing power to it, it is possible to synthesize two files with the same hash, as proven on https://shattered.io/[] in February 2017.
```
HonkingGoose, line 87:
```suggestion
If you dedicate several thousands of dollars' worth of computing power to it, it is possible to synthesize two files with the same hash, as proven on https://shattered.io/[] on February 2017.
```

I think you meant `on` here?

stuartpb: No. In English, events happen *in* months, not *on* them.

HonkingGoose: Oh wait, you're right, and I'm wrong, `in` is right in this case.... :+1:

ben:
I think we've got it! Thanks for the contribution!

PR #1554, opened 2020-11-03
Migrate to Asciidoctor 2.0.11

### LINKED ISSUE #1358: Proposal: use versions for other Ruby gems too.
I noticed that https://github.com/progit/progit2/blob/master/Gemfile only has versions listed for:
- gem 'asciidoctor', '1.5.6.2'
- gem 'asciidoctor-epub3', '~> 1.5.0.alpha.9'
- gem 'asciidoctor-pdf', '~> 1.5.0.beta.8'

But no versions are listed for:
- gem 'rake'
- gem 'json'
- gem 'awesome_print'
- gem 'coderay'
- gem 'pygments.rb'
- gem 'thread_safe'
- gem 'epubcheck'
- gem 'kindlegen'

It might be worth it to give known working versions to the other dependencies. That way Dependabot can help keep those up to date too. 

Doing this will also ensure that contributors/maintainers are on the same versions of everything, this might reduce instances of "Well it works on my computer, but not on Ben's.".

---

**Unable to install a working set of gems:**

I'm going to bump this issue as I tried to install Asciidoctor on my Windows 10 system and couldn't get a working set of dependencies installed with the `bundle install` command.

The problem is that the gems that are not given a version range will download the latest version of their gems, but those gems are not automatically compatible with the outdated version of Asciidoctor that we use. 

From memory, I think the issue was with the 'coderay' gem.

I would really like to get a _known working version range_ for all dependencies, instead of relying on the latest versions of some dependencies being compatible with the outdated version of Asciidoctor.

**Options to fix this:**

* My preferred method would be to implement PR #1373, so that we're back on a supported version of Asciidoctor again, and afterwards use a proper version range.
* Alternatively we use the version range that is currently in use on the maintainers workstation.

---

**History of the change:**

I also noticed the project used to have the gems locked to a specific version but that caused issues for **aollier** in getting a working tool-chain on his Fedora 25 system. #684

The argument given in favor of removal:
>It is unnecessary to track Gemfile.lock because it is platform dependent and it causes problems to build the book.

And a argument in favor of keeping the gem lock file:
>Well, the Gemfile.lock is supposed to insure a known-good combination of packages. But, right now, this leads to compilation errors indeed.

HonkingGoose:

## Changes

- Bump Asciidoctor from 1.5.6.2 to 2.0.11.
- Change style to new specifications.
- Pin dependencies.
- Fix unintended highlighting of text.
- Undo rake script on Markdown file.
- Undo rake script on Ruby file.

## Context

Fixes #1553
Fixes #1541 
Fixes #1358 
Fixes #1373

### DIFF
--- a/C-git-commands.asc
+++ b/C-git-commands.asc
@@ -41,7 +41,7 @@ In <<ch08-customizing-git#_keyword_expansion>> we showed how to set up smudge an
 
 Finally, basically the entirety of <<ch08-customizing-git#_git_config>> is dedicated to the command.
 
-[[_core_editor]]
+[[ch_core_editor]]
 ==== git config core.editor commands
 
 Accompanying the configuration instructions in <<ch01-getting-started#_editor>>, many editors can be set as follows:
@@ -97,7 +97,7 @@ To take a directory and turn it into a new Git repository so you can start versi
 
 We first introduce this in <<ch02-git-basics-chapter#_getting_a_repo>>, where we show creating a brand new repository to start working with.
 
-We talk briefly about how you can change the default branch name from ``master'' in <<ch03-git-branching#_remote_branches>>.
+We talk briefly about how you can change the default branch name from "`master`" in <<ch03-git-branching#_remote_branches>>.
 
 We use this command to create an empty bare repository for a server in <<ch04-git-on-the-server#_bare_repo>>.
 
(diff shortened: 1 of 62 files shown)


HonkingGoose:
I've squashed the changes that I and @jnavila made on #1373, and also merged #1541 into this branch.
I've also updated to the latest version of Asciidoctor (2.0.11) which seems to work just fine.

I manually migrated some missing legacy double quoted syntax to the modern syntax.

I think this is ready to go, but @jnavila can you please run that migration script again on this branch, and see what it catches?

@ben can you take a look at this and tell me what you think of it?

ben (approved):
LGTM :+1: You may merge when ready. :grin:

jnavila:
This was mostly a quoting issue. So OK for me.

HonkingGoose:
Woohoo! :partying_face: :balloon: :fireworks: :sparkles:

HonkingGoose:
By the way the project board that @jnavila made on the `progit2` repo can be closed/renamed/deleted as we no longer need it for the migration. :wink:

Issue #605, opened 2016-04-27
booyah/master is remote-tracking branch, instead of remote branch.


YueLinHo:
`book\03-git-branching\sections\remote-branches.asc`
Line: 27
```
If you run `git clone -o booyah` instead, 
then you will have `booyah/master` as your default remote branch.(((origin)))
```

Change to?
```
If you run `git clone -o booyah` instead, 
then you will have `booyah/master` as your default remote-tracking branch.(((origin)))
```
remote branch -> remote-tracking branch

YueLinHo:
@ben I am going to fix this issue by myself. Could you please tell me which font fact is in that figure?

ben:
Looks like the font in the Sketch file is Menlo. We use the Regular and Bold variants.

YueLinHo:
Looks like Menlo is Mac only!!!?
Is it good for you that using to **Bitstream Vera Sans Mono** for Menlo? (Fallback)
Download it [here](http://ftp.gnome.org/pub/GNOME/sources/ttf-bitstream-vera/1.10/) (The ttf-bitstream-vera-1.10.zip file)

cf. 
(1) Wike - [Menlo (typeface)](https://en.wikipedia.org/wiki/Menlo_(typeface))
(2) The [answer](https://superuser.com/a/59605/585060) for "Any good way to get the new Menlo font on Windows?" on superuser
(3) [ Comparison: Bitstream vs DejaVu vs Menlo](https://blog.fnurl.se/2011/06/22/comparison-bitstream-vs-dejavu-vs-menlo/)

thernstig:
@YueLinHo also noticed this now, did you get further with fixing this?

PR #1558, opened 2020-11-10
Update remote-branches.asc

Leative:

## Changes

- adapted terms and explanations with regard to the differences between remote branches, remote-tracking branches and tracking branches

## Context
The document differentiates between remote branches, remote-tracking branches and tracking branches. I tried to fix some ambiguities which could, in my opinion, be misleading.

Related #605

### DIFF
--- a/book/03-git-branching/sections/remote-branches.asc
+++ b/book/03-git-branching/sections/remote-branches.asc
@@ -24,7 +24,7 @@ Git also gives you your own local `master` branch starting at the same place as 
 ====
 Just like the branch name "`master`" does not have any special meaning in Git, neither does "`origin`".
 While "`master`" is the default name for a starting branch when you run `git init` which is the only reason it's widely used, "`origin`" is the default name for a remote when you run `git clone`.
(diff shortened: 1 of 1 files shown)


slawomirbudzin53 (approved):

ben:
Remember, Pro Git isn't a reference manual. It's meant to be a readable introduction to Git for people who have never used it before. So adding technical correctness at the expense of accessibility to newbies, _especially_ in the first chapter where we talk about branches, doesn't fit. 

Also worth noting here: "remote branch" and "remote-tracking branch" don't have an _official_ definition per se. They're just refs to Git, so you can define them however you want. In this book we tried to define "remote branch" as "the local ref that represents the state of the ref on the remote," but we may have slipped a bit, and if that's the case it's worth fixing.

If you'd like to take another stab at this, I'll happily review it. I'd love to resolve some of the ambiguity we have here, but not at the expense of the beginner's experience.

revmarcuspeter-cyber (approved):

PR #1561, opened 2020-11-16
Escape tilde to fix HEAD~~~ presented as HEAD~

ixth:

## Changes

- Fixed \`\~\~\~\` presented as subscript tilde.

## Context

I noticed that \~\~\~ is presented as subscript tilde on git-scm site for some reason:

Screenshot 2020-11-16 at 12 43 01

### DIFF
--- a/book/07-git-tools/sections/revision-selection.asc
+++ b/book/07-git-tools/sections/revision-selection.asc
@@ -282,7 +282,7 @@ Date:   Fri Nov 7 13:47:59 2008 -0500
     Ignore *.gem
 ----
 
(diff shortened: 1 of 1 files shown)


HonkingGoose:
I've checked the latest ProGit2 PDF release (2.1.272) and that also displays the subscript tilde.

I've not tested this branch with a build of my own, so I don't know if the proposed fix is working correctly or not.

ixth:
I've built and checked pdf, and it looks just fine:

ben:
Rad, thanks!

PR #1563, opened 2020-11-21
Fix formatting issue

yugaego:
Asciidoctor documentation suggests solution to properly render a possessive monospaced phrase by using [unconstrained formatting](https://asciidoctor.org/docs/user-manual/#unconstrained-formatting-edge-cases).
However, this approach does not work when one more monospaced phrase follows [in the same paragraph](https://gist.github.com/yugaego/edcef321c3e4a7c4f59d7756d7ef4042).
The simplest workaround was suggested by the Asciidoctor project lead Dan Allen.

**Before:**
PDF
before-pdf

HTML
before-html

EPUB
before-epub

**After:**
PDF
after-pdf

HTML
after-html

EPUB
after-epub

### DIFF
--- a/book/07-git-tools/sections/advanced-merging.asc
+++ b/book/07-git-tools/sections/advanced-merging.asc
@@ -580,7 +580,7 @@ The history with the revert commit looks like this:
 .History after `git revert -m 1`
 image::images/undomerge-revert.png[History after `git revert -m 1`]
 
-The new commit `^M` has exactly the same contents as `C6`, so starting from here it's as if the merge never happened, except that the now-unmerged commits are still in `HEAD`'s history.
+The new commit `^M` has exactly the same contents as `C6`, so starting from here it's as if the merge never happened, except that the now-unmerged commits are still in `HEAD`’s history.
 Git will get confused if you try to merge `topic` into `master` again:
 
 [source,console]


ben:
Kind of sad, I really don't like to put non-ASCII characters into the source code like this. Is there no other way?

jnavila:
Come on, the world is UTF-8 nowadays.

Personal taste, but I strongly dislike these “macro” things that can replace characters. They always end up colliding with perfectly admissible use of the reserved sequences, and require more ugly escaping.

yugaego:
> Is there no other way?

These were the other options suggested on Gitter earlier:

> Abel Salgado Romero @abelsromero Nov 21 22:35
> You could just passtrough "Text with `HEAD`pass:[']s reference."
> Or use the special attribute {apos} instead


So, I have tested the options, and the following ones gave expected results (screenshots are shown below):

```asciidoctor
`HEAD`pass:[’]s with pass.

`HEAD`{rsquo}s curly brackets.

`HEAD`&#8217;s html entity. 

```

Note that the `right single quote` (not apostrophe) would have to be set explicitly to be consistent with the rest of the text. 

So, to me **inserting the quote directly seems to be a more straightforward option** in this case.

*Some code context*
https://github.com/asciidoctor/asciidoctor/blob/dd843d59b1e49f29d68386bfdf0438ce11b42577/lib/asciidoctor.rb#L393
https://github.com/asciidoctor/asciidoctor/blob/dd843d59b1e49f29d68386bfdf0438ce11b42577/lib/asciidoctor.rb#L507


*HTML*
options-html
*EPUB*
options-epub
*PDF*
options-pdf

ben:
> Personal taste, but I strongly dislike these “macro” things that can replace characters.

I mean, I'm kind of with you here. If it were a 😛 emoji, I would totally advocate to include the literal. But punctuation is a weird thing where semantics and display really are separate things (i.e. German-style „lower quotation marks”, but also perhaps a particular format wants `&lquo;` entities). 

> to me **inserting the quote directly seems to be a more straightforward option** in this case.

Okay. In this case it seems like the right thing to do, but I worry that translation projects might end up with rendering issues because this character wasn't re-typed exactly correctly.

Thanks for catching and fixing this!

jnavila:
:bow:

PR #1564, opened 2020-11-24
Introduce test of html and epub

jnavila:
This is an indirect way of checking the structure of the book, the
asciidoc syntax and the external links.

## Changes

-


jnavila:
There are some errors in the source files that need fixup. Already done
 
* proper escaping of `*` characters
* proper quoting in images'alternate texts

What's still needed (at least) is anchor naming.

Gemfile
```diff
@@ -12,4 +12,5 @@ gem 'asciidoctor-pdf', '1.5.3'
 gem 'coderay', '1.1.3'
 gem 'pygments.rb', '1.2.1'
 gem 'thread_safe', '0.3.6'
-gem 'epubcheck', '3.0.1'
+gem 'epubcheck-ruby'
+gem 'html-proofer'
```
HonkingGoose, line 16:
```suggestion
gem 'epubcheck-ruby', '4.2.4.0'
gem 'html-proofer', '3.17.2'
```

I think it's a good idea to pin these dependencies as well, just like we're doing for all the other dependencies.

Can you check if this is right? I've grabbed the latest versions from the rubygems.org site.

jnavila: It is right. But I'm not sure that we want to rely on a pinned version of our checkers. Another point is the number of errors coming from these checks...

book/03-git-branching/sections/workflows.asc
```diff
@@ -23,7 +23,7 @@ It's generally easier to think about them as work silos, where sets of commits g
 
 [[lrbranch_b]]
 .A "`silo`" view of progressive-stability branching
-image::images/lr-branches-2.png[A "`silo`" view of progressive-stability branching]
+image::images/lr-branches-2.png[A “silo” view of progressive-stability branching']
```
ben, line 26: This is weird. It recommended you freeze the curly quotes? I'd rather they just be straight quotes if we can't use the ``` "`backtick`" ``` syntax we use elsewhere.

Also, what's with the trailing single quote?

book/06-github/sections/2-contributing.asc
```diff
@@ -123,7 +123,7 @@ To https://github.com/tonychacon/blink
 
 Now if we go back to our fork on GitHub, we can see that GitHub noticed that we pushed a new topic branch up and presents us with a big green button to check out our changes and open a Pull Request to the original project.
 
-You can alternatively go to the "`Branches`" page at `https://github.com/<user>/<project>/branches` to locate your branch and open a new Pull Request from there.
+You can alternatively go to the "`Branches`" page at `\https://github.com/<user>/<project>/branches` to locate your branch and open a new Pull Request from there.
```
ben, line 126: Is ``` `h ``` some kind of special code in Asciidoc? What was the error code here?

ben (commented):
So many ❤️s for this concept, I'm just a little mystified by some of the changes these linters are recommending.

book/03-git-branching/sections/workflows.asc
jnavila: I don't know if it's a feature or a bug of asciidoctor, but the backticks macro is not converted in the `alt` attributes of images, leading to a breakage of xhtml syntax.

Single quoting may work and remain as is, but it's a semantic change. And anyway, is an alternate text that just replicates the caption really needed?

book/06-github/sections/2-contributing.asc
jnavila: With the new syntax of asciidoctor, the backticks no longer convey a semantic format, but are just mere font changes. So they no longer escape other formating inside the ticks. Here the "https://" form is detected and converted into a link, which brings three issues:

* is a url without the trailing `[]` supposed to be converted
* This is not a real URL so converting is undesirable
* Even if we let it converted, the conversion has a bug which does not handle the < and > entities correctly and produces invalid xhtml code.

Putting a backslash just disables URL scheme detection and lets the backticks do the font change.

jnavila: Replying to first point: this is a feature called "autolink".

book/03-git-branching/sections/workflows.asc
jnavila: For ref, https://github.com/asciidoctor/asciidoctor/issues/3864

book/06-github/sections/2-contributing.asc
jnavila: For Ref: Please see https://github.com/asciidoctor/asciidoctor/issues/3864

book/03-git-branching/sections/workflows.asc
jnavila: OK, for the trailing single quote.

Would you prefer converting:

* “ → '
* ” → ' 

in alternative texts?

ben: Yeah, this is a fair point. Ideally we'd have semantic alt-text that actually helps in a sight-impaired scenario. I don't remember why we did it this way.

I guess it's fine, though.

book/06-github/sections/2-contributing.asc
ben: Awesome, thanks for the info.

ben (approved):

PR #1567, opened 2020-11-29
Small rewording change inside Rewriting History

tiffany-taylor:

## Changes
A small typo fix within the Rewriting History section, changing "they" to "that" in the below sentence:
> To modify a commit that is farther back in your history, you must move to more complex tools. Git doesn’t have a modify-history tool, but you can use the rebase tool to rebase a series of commits onto the HEAD they were originally based on instead of moving them to another one.


tiffany-taylor:
Well, this is a great example of making sure my changes are correct before submitting. I reread the sentence again, and I can see  how the sentence makes sense in context, but the wording is a little clumsy, in my opinion. Possibly "that they were originally based on" would be better than my suggested change, or what is currently present.

ben:
Yeah, this could use a rewrite. This definitely makes it better though. Thanks!

PR #1569, opened 2020-11-30
Use -L option in curl command to follow redirect

drichardson:

## Changes

Use -L option in curl command to follow redirect returned by www.kernel.org. Otherwise, the file you are testing with is not the ~5MB git.tgz, but a 168 byte HTML page informing you about the redirect.

## Context

Everything under https://www.kernel.org/pub/ seems to 301 (moved permanently) redirect to https://mirrors.edge.kernel.org/pub/. So the other option is to change the URL instead of adding the `-L` curl option. However, the www.kernel.org homepage still links to them as https://www.kernel.org/pub/, so it isn't clear to me what the most stable URL is, which is why I went with the -L.

### DIFF
--- a/book/10-git-internals/sections/maintenance.asc
+++ b/book/10-git-internals/sections/maintenance.asc
@@ -192,7 +192,7 @@ First, add a large object to your history:
 
 [source,console]
 ----
-$ curl https://www.kernel.org/pub/software/scm/git/git-2.1.0.tar.gz > git.tgz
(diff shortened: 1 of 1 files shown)


ben:
Hmmm. Would it make more sense to update that URL to something more current? Do the newer releases have stable URLs?

drichardson:
I don't know. Since the content of the tarball is irrelevant in that example anyway, maybe you should just link to something else under your control (e.g. something on github) so you don't have to deal with the possibility of it changing.

ben:
Yeah, I guess you're right. Thanks!

PR #1583, opened 2020-12-23
Make wording in refs.asc clearer

vegerot:
When reading this line I was very confused.  It didn't sound like it made sense.  So I checked out a remote ref, ran `cat .git/HEAD`, and lo and behold it did "point" to the same thing that the remote branch pointed to.

I played around with it a little bit more and realized the difference between checking out a local and remote reference is that checking out a local reference sets `.git/HEAD` to a SYMBOLIC ref, while checking out a remote reference sets it to the exact commit.

I think the current language is *technically* correct, because "Git won't point HEAD at one" means it won't point HEAD to the remote reference itself (since "one" means the remote reference, not the commit).  So I'm not saying the current wording is bad, just that it confused me, so it might confuse others

Also, when you say "you'll never update it" and "read-only" of course you are referring to the remote reference, but a naive reader like myself may thing you're referring to HEAD.  And if I checkout a remote reference then commit a change I will update HEAD

## Changes

-

### DIFF
--- a/book/10-git-internals/sections/refs.asc
+++ b/book/10-git-internals/sections/refs.asc
@@ -205,5 +205,5 @@ ca82a6dff817ec66f44342007202690a93763949
 ----
 
 Remote references differ from branches (`refs/heads` references) mainly in that they're considered read-only.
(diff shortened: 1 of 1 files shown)


ben:
Yeah, that's better. Thanks!

PR #1586, opened 2020-12-30
Replace 3 instances of "VCS systems" with "VCSs"

SherryHietala:

## Changes

- The redundant phrase "VCS systems" should be replaced with "VCSs" in three locations in the text (sections 1.1, 2.1, 2.2).

## Context

"VCSs" as a plural abbreviation is established in the Introduction and used elsewhere in the book.


ben:
Nice, thanks!

PR #1603, opened 2021-02-09
Fix lingering Markdown for git-send-email.io

noelleleigh:

## Changes

- Fix the link to <https://git-send-email.io> to use AsciiDoc syntax instead of Markdown.

## Context

Builds on #1489

### DIFF
--- a/book/05-distributed-git/sections/contributing.asc
+++ b/book/05-distributed-git/sections/contributing.asc
@@ -787,7 +787,7 @@ Result: OK
 
 [TIP]
 ====
-For help on configuring your system and email, more tips and tricks, and a sandbox to send a trial patch via email, go to https://git-send-email.io/[git-send-email.io].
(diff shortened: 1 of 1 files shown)


HonkingGoose:
@noelleleigh Good catch! I got my Markdown and my Asciidoctor link format mixed up... 🙃 

I like the window `_blank` property you're using with the caret!

Quote from [Asciidoctor writers guide, links and images](https://asciidoctor.org/docs/asciidoc-writers-guide/#links-and-images):

> Since `_blank` is the most common window name, we’ve introduced shorthand for it. Just end the link text with a caret (`^`):

```
Let's view the raw HTML of the link:view-source:asciidoctor.org[Asciidoctor homepage^].
```

Maybe we should use that by default, so that the ProGit 2 book stays open in the original tab whenever you click on an external link.

PR #1611, opened 2021-02-15
Partial (and shallow) clone note

### LINKED ISSUE #1597: Proposal: Partial clone
I proposed to add discussion about partial clone (blobless, treeless, and shallow clones) to the book.

For the text, it can be inspired from [this GitHub blog post](https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone/).

In summary:

When you want to clone a repository, but you don't want full clone (for example because you're on metered internet connection), Git provides three ways to get partial clone:
  1. `git clone --filter=blob:none <url>` - blobless clone. Trees and commits are downloaded, but blobs are downloaded on-demand after checkout. Recommended for developers
  2. `git clone --filter=tree:0 <url>` - treeless clone. Only commits are downloaded. Blobs and trees are downloaded on-demand. Recommended for CI builds that needs access to commit history
  3. `git clone --depth=1 <url>` - shallow clone. Download commit history until specified depth level (roughly corresponds to number of commits), in this case 1 commit. Recommended for CI builds.

bagasme:

## Changes

Add note about partial and shallow clone, with external link describing
them in detail. The note is in Chapter 2 - Getting a repository

## Context

Fixes #1597

### DIFF
--- a/book/02-git-basics/sections/getting-a-repository.asc
+++ b/book/02-git-basics/sections/getting-a-repository.asc
@@ -85,11 +85,14 @@ That command does the same thing as the previous one, but the target directory i
 [NOTE]
 .Partial and shallow clone
 ====
-Git version 2.22 and later introduce `--filter` option to `git clone`, which creates _partial clone_. Beside that, you can also get _shallow clone_ by specify `--depth` option.
+Git version 2.22 and later introduce `--filter` option to `git clone`, which creates _partial clone_.
+Beside that, you can also get _shallow clone_ by specify `--depth` option.
 
-While partial and shallow clones can be useful on some situations (such as when cloning extremely large repository and cloning over metered connection), these come with their own tradeoffs that break at least one expectation from normal distribute nature of Git. If you aren't willing to make these tradeoffs, just do normal (full) clone (`git clone` without `--filter` or `--depth` options).
+While partial and shallow clones can be useful on some situations (such as when cloning extremely large repository and cloning over metered connection), these come with their own tradeoffs that break at least one expectation from normal distribute nature of Git.
+If you aren't willing to make these tradeoffs, just do normal (full) clone (`git clone` without `--filter` or `--depth` options).
 
-We won't discuss further about partial and shallow clones, as these are for advanced users. Please see GitHub blog post at https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone/ and GitLab documentation at https://docs.gitlab.com/ee/topics/git/partial_clone.html for complete discussion about these.
+We won't discuss further about partial and shallow clones, as these are for advanced users.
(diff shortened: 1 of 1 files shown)


HonkingGoose (changes requested):
I have more suggestions coming up, but first follow the "one sentence per line" structure.

bagasme:
review ping

book/02-git-basics/sections/getting-a-repository.asc
```diff
@@ -82,6 +82,19 @@ $ git clone https://github.com/libgit2/libgit2 mylibgit
...
+.Partial and shallow clone
+====
+Git version 2.22 and later introduce `--filter` option to `git clone`, which creates _partial clone_.
+Beside that, you can also get _shallow clone_ by specify `--depth` option.
+
+While partial and shallow clones can be useful on some situations (such as when cloning extremely large repository and cloning over metered connection), these come with their own tradeoffs that break at least one expectation from normal distribute nature of Git.
+If you aren't willing to make these tradeoffs, just do normal (full) clone (`git clone` without `--filter` or `--depth` options).
+
+We won't discuss further about partial and shallow clones, as these are for advanced users.
+Please see GitHub blog post at https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone/ and GitLab documentation at https://docs.gitlab.com/ee/topics/git/partial_clone.html for complete discussion about these.
```
HonkingGoose, line 95:
```suggestion
Git version 2.22 and later introduce the `--filter` option to `git clone`, which creates a _partial clone_.
Beside that, you can also get a _shallow clone_ by specifying the `--depth` option.

Partial and shallow clones can be useful in some situations, like when you're cloning an extremely large repository and cloning over a metered connection.
The benefit of doing a partial or shallow clone is that the clone takes up less size on disk.
The drawback is that a partial or shallow clone breaks the normal behavior of Git.

If you don't want to deal with this complexity, just do a normal (full) clone (`git clone` without `--filter` or `--depth` options).

We won't discuss partial and shallow clones in depth here, as these clones are for advanced users.
Please see GitHub blog post at https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone/ and GitLab documentation at https://docs.gitlab.com/ee/topics/git/partial_clone.html for complete discussion about these.
```

How about something like this?

bagasme: @HonkingGoose looks ok

bagasme:
review ping

ben:
I have two minds on this PR. On the one hand, it's well-written, and valuable info for the reader.

But does it belong in chapter 2? This is where a reader who has never worked with Git before gets their first taste. We deliberately prioritize that first-time experience in this chapter over things like absolute correctness or deep-dives into topics. It seems to me that this content would be a better fit as a section in chapter 7, where there's some room to dig into some common use cases.

Would you want to take a swing at that larger chunk of content?

bagasme:
@ben ok

revmarcuspeter-cyber (approved):

PR #1621, opened 2021-03-01
Index must be only in pdf

SergeiKuznetsov:

 ## Changes
 * Fixed include directive
 * Used level 1 for Index section
 see https://docs.asciidoctor.org/asciidoc/latest/sections/
 * Deleted unused file `book/index.asc`


Anezdoank04 (approved):

ben:
Can you explain why this needs to change? Why should we not ship the index with the HTML format?

SergeiKuznetsov:
> Can you explain why this needs to change? Why should we not ship the index with the HTML format?

The built-in HTML5 converter in Asciidoctor does not generate an index.
See [docs](https://docs.asciidoctor.org/asciidoc/latest/sections/)

ben:
Alright. Let's give it a shot.

Issue #1627, opened 2021-03-06
Git for Windows branch name contain double quotes


wtask:
We have new developer who creates the branches like this
```
bla-bla-bla_again_bla_bla_bla_"something_special"
```
So if I try to fetch that branch under windows, I'll get an error (unable to update local ref) related to lock file creation due to double quotes is a part of branch name and also the lock file name under Windows.

`git version 2.30.1.windows.1`

HonkingGoose:
Hi @wtask! :wave:

This repository is about the _ProGit2 book_, that helps readers to learn about Git.

It seems you're trying to report a problem with the _Git for Windows_ program.
You can report problems with Git for Windows at this link: https://github.com/git-for-windows/git/issues

wtask:
@HonkingGoose sorry

HonkingGoose:
@wtask Don't worry about it. It's easy to end up in the wrong place when you're searching for help. I've opened issues in the wrong place as well.... 😄 

I've opened a PR to make the wording on our issue select screen clearer, because it does say: "Open issue here", which does sound like you are in the right place on the ProGit2 repository for some of the issues. I apologize for my bad wording on these items! 😉

PR #1628, opened 2021-03-06
Improve wording on issue select screen

HonkingGoose:

## Changes

- Change word `here` -> `there`

## Context

I think the current wording is confusing to newcomers to the ProGit2 repository: see #1627.
The user probably meant to post at the Git for Windows repository, but somehow still ended up here. 😄 
We can all take a wrong turn somewhere, but we should still ensure that we're clear on where to go instead.

Feel free to suggest further improvements to the text, maybe we should be really explicit and say:
`Click on the "open" button next to this text to go to the right place to report your problem.`?


ben:
😁👍🏼

PR #1650, opened 2021-05-21
Update C-git-commands.asc with a new TextEdit command

pibion:

## Changes

- Update C-git-commands.asc with a new TextEdit command

## Context

The existing git config instructions for TextEdit on Mac would hang after closing the TextEdit program. The instructions outlined in https://ericasadun.com/2018/07/12/using-textedit-as-your-git-editor/ worked and I propose updating the book with her solution.

### DIFF
--- a/C-git-commands.asc
+++ b/C-git-commands.asc
@@ -62,7 +62,7 @@ Accompanying the configuration instructions in <<ch01-getting-started#_editor>>,
 |Scratch (Linux)|`git config --global core.editor "scratch-text-editor"`
 |Sublime Text (macOS) |`git config --global core.editor "/Applications/Sublime\ Text.app/Contents/SharedSupport/bin/subl --new-window --wait"`
 |Sublime Text (Windows 64-bit) |`git config --global core.editor "'C:\Program Files\Sublime Text 3\sublime_text.exe' -w"` (Also see note below)
-|TextEdit (macOS)|`git config --global core.editor "open --wait-apps --new -e"`
+|TextEdit (macOS)|`git config --global core.editor "open -W -n"`
 |Textmate |`git config --global core.editor "mate -w"`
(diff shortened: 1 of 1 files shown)


C-git-commands.asc
```diff
@@ -62,7 +62,7 @@ Accompanying the configuration instructions in <<ch01-getting-started#_editor>>,
 |Scratch (Linux)|`git config --global core.editor "scratch-text-editor"`
 |Sublime Text (macOS) |`git config --global core.editor "/Applications/Sublime\ Text.app/Contents/SharedSupport/bin/subl --new-window --wait"`
 |Sublime Text (Windows 64-bit) |`git config --global core.editor "'C:\Program Files\Sublime Text 3\sublime_text.exe' -w"` (Also see note below)
-|TextEdit (macOS)|`git config --global core.editor "open --wait-apps --new -e"`
+|TextEdit (macOS)|`git config --global core.editor "open -W -n"`
```
ben, line 65: Three things:

1. The `-W` flag is short for `--wait-apps`, and `-n` is short for `--new`. When we're writing something that will go into a file and be forgotten until we're trying to figure out why it broke, we want to optimize for readability, so I'd prefer we keep the long version of the flags here.
2. The `-e` flag specifies that the file should be opened in TextEdit. Without it, you'll get whatever app registers itself as the default app for plain text files. We're specifically trying to get TextEdit here, so removing the `-e` flag just makes this less specific.
3. I just tested both variations of this, and they behave the same? You have to actually command-Q the app to make `open` recognize that it's done, but otherwise I don't see a difference.

pibion: Keeping the long version of the flags makes sense.

I'm not sure why we saw a different behavior on our system between these two commands, and I'll see if we can re-test to confirm that we weren't imagining things.

sabbir999Q (approved):

revmarcuspeter-cyber (approved):

revmarcuspeter-cyber (approved):

PR #1655, opened 2021-06-15
Enable .mobi (Kindle) format again

### LINKED ISSUE #1528: Discussion: does the ProGit2 project want to keep providing Kindle files?
### Do we want to keep providing `.mobi` files?

The ProGit2 build system broke because the `kindlegen` dependency is no longer available.

I have not seen an issue report here or on the git-scm repo complaining that there are no files for their Kindle devices.
Maybe those readers just use the `.epub` `.pdf` files instead, or they just read the book on the `git-scm` site itself?

Maybe we can just stop providing `.mobi` files altogether?

### References:

- The problem with the build breaking was raised in #1496.
- The broken build was fixed with pull #1497 which just drops the `kindlegen` dependency, and stops building `.mobi` files.
- The issue was also raised with the `git-scm` project at https://github.com/git/git-scm.com/issues/1498.

HonkingGoose:

## Changes

- Enable `.mobi` (Kindle) format again
- Add references to mobi files back in README
- Add mobi rake build command to README

## Context

Fixes #1528

`asciidoctor-epub3` has the Kindlegen dependency in it again, so we can enable mobi builds again. (https://github.com/progit/progit2/pull/1646#issuecomment-828968577)

Please check if this change alone is enough to get things working.

@slonopotamus @jnavila can you please check my work?

### DIFF
--- a/README.asc
+++ b/README.asc
@@ -15,8 +15,7 @@ See link:TRANSLATING.md[the translating document] for more information.
 == How To Generate the Book
 
 You can generate the e-book files manually with Asciidoctor.
-We used to be able to build .mobi files (Kindle), but cannot do so now, see #1496 for more information.
-If you run the following you _may_ actually get HTML, Epub and PDF output files:
+If you run the following you _may_ actually get HTML, Epub, Mobi and PDF output files:
 
 ----
 $ bundle install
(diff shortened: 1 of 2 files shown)


jnavila (approved):
Straightforward. Thank you!

jnavila:
In fact, the mobi file is missing in the cleaning task!

HonkingGoose:
@jnavila can you point me to what went wrong?

HonkingGoose:
I think the refactoring PR #1598 made some more changes, that just inserting the old code is no longer working properly....

PR #1662, opened 2021-06-23
Revert "Enable .mobi (Kindle) format again"

HonkingGoose:
Reverts progit/progit2#1655

This PR is breaking the build on the `master` branch, so we're not releasing new files for any merged PRs.
I'd rather revert my bad change, and let somebody else try again.


HonkingGoose:
Root cause of the broken build might be that travis CI has shut down their old domain, we don't seem to have any functional builds anymore... 😄

PR #1667, opened 2021-06-27
Set up GitHub actions for CI and release builds

### LINKED ISSUE #1566: Migrate away from travis-ci.org
travis-ci.**org** is going to be shut down on **December, 31**:

https://docs.travis-ci.com/user/migrate/open-source-repository-migration#frequently-asked-questions

> Q. When will the migration from travis-ci.org to travis-ci.com be completed?
> A. In an effort to ensure that all of our users - whether you build open-source, public or private repositories - receive regular feature updates, security patches and UX/UI enhancements, we are announcing that travis-ci.org will be officially closed down completely no later than December 31st, 2020, allowing us to focus all our efforts on bringing new features and fixes to travis-ci.com and all of our awesome users like yourself on the travis-ci.com domain.

> Q. What will happen to travis-ci.org after December 31st, 2020?
> A. Travis-ci.org will be switched to a read-only platform, allowing you to see your jobs build history from all repositories previously connected to travis-ci.org.

Also, note that travis-ci.**com**, while being the most logical place to move to, has **build time limits** for public repositories:

https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing

> For those of you who have been building on public repositories (on travis-ci.com, with no paid subscription), we will upgrade you to our trial (free) plan with a 10K credit allotment (which allows around 1000 minutes in a Linux environment).

I saw that multiple projects (for example, major part of Asciidoctor ecosystem) moved from Travis to GitHub Actions in last month or two, though this is not a single option available nowadays.

ben:

## Changes

This sets up GitHub actions to be our book-build provider. Fixes #1566.

- [x] Action to build artifacts on PRs and merges to `master`
- [x] Action to tag on merges to `master`
- [x] Action to create a release when a new tag is pushed


.github/workflows/tag-on-merge.yml
```diff
@@ -0,0 +1,22 @@
...
+      id: compute-tag
+      run: |
+        echo Computing next tag number
+        LASTPATCH=$(git describe --tags | cut -d- -f1 | cut -d. -f3)
+        echo "::set-output name=tagname::$(($LASTPATCH+1))"
+    - name: Create Tag
+      uses: negz/create-tag@v1
+      with:
+        version: ${{ steps.compute-tag.outputs.tagname }}
+        token: ${{ secrets.GITHUB_TOKEN }}
```
HonkingGoose, line 22:
```suggestion
        token: ${{ secrets.GITHUB_TOKEN }}

```

Maybe add a newline here? The GitHub preview is complaining about the EOL marker, I think?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,20 @@
...
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Ruby
+      uses: ruby/setup-ruby@v1
+      with:
+        ruby-version: 2.7
+        bundler-cache: true # runs 'bundle install' and caches installed gems automatically
+    - name: Run tests
```
HonkingGoose, line 19:
```suggestion
    - name: Build book
```

This is more descriptive than just the generic "Run tests" name. 😉

.github/workflows/tag-on-merge.yml
ben: Got it! New laptop, editor config isn't quite there yet.

.github/workflows/ci.yml
ben: 👍🏼

ben:
@HonkingGoose how does this look? I'm afraid it's hard to test at this point without merging.

.github/workflows/release-on-tag.yml
```diff
@@ -0,0 +1,30 @@
+name: Tag on merge to master
+
+on:
+  push:
+    tags: [ '*' ]
```
HonkingGoose, line 5: I think using `'*'` is okay, unless we have any tags with a `/` in them? I think our tags follow `x.x.x` pattern as far as I can tell?

From the [GitHub docs, GitHub Actions Filter pattern cheat sheet](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet):

> You can use special characters in path, branch, and tag filters.
>
> - `*`: Matches zero or more characters, but does not match the `/` character. For example, `Octo*` matches `Octocat`.
> - `**`: Matches zero or more of any character.

We're correctly quoting the wildcard/filter pattern, according to the docs:

The characters `*`, `[`, and `!` are special characters in YAML. If you start a pattern with `*`, `[`, or `!`, you must enclose the pattern in quotes.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,22 @@
+name: CI
```
HonkingGoose, line 1: Is `name: CI` the best name for this workflow? Would `build.yml` and `name: Build` be clearer?

.github/workflows/release-on-tag.yml
```diff
@@ -0,0 +1,30 @@
+name: Tag on merge to master
```
HonkingGoose, line 1:
```suggestion
name: Release on tag on master
```

This might be clearer? It would be similar to the current filename: `release-on-tag.yml`.

.github/workflows/tag-on-merge.yml
```diff
@@ -0,0 +1,25 @@
...
+
+    - name: Compute tag name
+      id: compute-tag
+      run: |
+        echo Computing next tag number
+        LASTPATCH=$(git describe --tags | cut -d- -f1 | cut -d. -f3)
+        PATCH=$(($LASTPATCH+1))
+        echo "::set-output name=tagname::2.1.${PATCH})"
+
+    - name: Create Tag
```
HonkingGoose, line 21:
```suggestion
    - name: Create tag
```

We don't use a lowercase `tag` in the other `name:` in this file, so should probably be consistent across both names. 😉

HonkingGoose (commented):
I think it looks OK, as far as I can tell. Some small niggles.

HonkingGoose:
> I'm afraid it's hard to test at this point without merging.

I agree, it's probably easiest to debug while running it for real, at least then you can see the output and tune things that are not working properly. 😄 

You could copy/clone this repository into a new "throwaway/trial" repository, and run the actions on there?

.github/workflows/release-on-tag.yml
ben: Pretty sure this is okay? We're really only interested in automatic tags, which are always `2.1.xxx`, so this should work just fine for that.

.github/workflows/ci.yml
ben: OK, I'm cool with that. I'm also going to remove the run on `master` because we're doing the same thing with the release-on-tag workflow.

.github/workflows/release-on-tag.yml
ben: Whoops, copy-paste error.

.github/workflows/release-on-tag.yml
HonkingGoose: I think it's OK as well, but just wanted to verify. 😉

.github/workflows/tag-on-merge.yml
```diff
@@ -0,0 +1,31 @@
...
+jobs:
+  tag:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Compute tag name
+      id: compute-tag
+      run: |
+        echo Computing next tag number
```
HonkingGoose, line 16:
```suggestion
        echo "Computing next tag number"
```

Not sure if this is needed? I notice we use the `echo "..."` pattern 3 lines down?

ben: I'm not totally sure of the difference here? But [the runs](https://github.com/ben/progit2-test/runs/2925691880?check_suite_focus=true#step:3:7) work okay.

HonkingGoose: It seems to work OK on your `progit-2-test` repository.

Let's leave it as-is then. 😄

HonkingGoose:
I notice we have some differences between this PR, the `progit2-test/master` branch, and the `progit2-test/ben-github-actions` branch.

Are you sure this PR contains all the parts you need? I saw you did something with the git fetch depth on the `progit2-test/master` branch for example.

HonkingGoose:
@ben Might want to put `Fixes #1566` in the top post, so that that issue gets closed when you merge this PR.

See issue: #1566.

HonkingGoose:
By the way the `.mobi` files are not getting released properly on your test repo. Maybe there's a problem with the link to the file?

According to: https://github.com/ben/progit2-test/actions/runs/976418220

```
tag
Artifact pattern :./progit.mobi did not match any files
```

ben:
> `.mobi` files are not getting released properly

I think this is because kindlegen isn't there? I'm completely fried from driving all day and this 112° heat. I'll take another look in the morning.

Also, sadly, creating a tag with the API doesn't cause a `push` event, so I'll have to collapse the "tag on merge to master" and "create release on tag" jobs into one. Too bad.

HonkingGoose:
> I think this is because kindlegen isn't there? I'm completely fried from driving all day and this 112° heat. I'll take another look in the morning.

The `asciidoctor-epub3` package "includes" Kindlegen by downloading it from someplace else, quote from latest release notes for the package (which we already merged):

> Possibly even more exciting feature of this release is that Kindlegen is back! Even though Amazon dropped Kindlegen from its webservers, what was put on the Internets stays there forever. Thanks to The Wayback Machine that remembers (almost) everything, we can use it to download Kindlegen binaries.

https://github.com/progit/progit2/pull/1646#issuecomment-828968577

Maybe this download is not working properly on CI systems? Or could be caused by something else. 😄

ben:
Boom, [here's a release](https://github.com/ben/progit2-test/releases/tag/2.1.1005) created by these actions. It doesn't include the .mobi, because it says it [can't find kindlegen](https://github.com/ben/progit2-test/runs/2932279997?check_suite_focus=true#step:5:14). Not really sure what's going on there.

ben:
I'm tempted to merge this just so we can get releases flowing again, then follow up with a kindlegen fix and cleaning up our Travis mess.

HonkingGoose:
> I'm tempted to merge this just so we can get releases flowing again, then follow up with a kindlegen fix and cleaning up our Travis mess.

Sounds good!

PR #1677, opened 2021-07-06
Fix descriptions of untracked files

bkline:
## Licensing

## Changes

- Make it clear that the set of tracked files includes more than those in the latest commit.


ben:
Yeah, this looks better. Thanks!

PR #1682, opened 2021-07-09
Link to @jnavila's GitHub profile

HonkingGoose:

## Changes

- Make the name `@jnavila` a link to their GitHub profile

## Context

This makes it easier to find out who you need to ping when you need help setting up the EPUB publication chain.

Maybe this line can be removed? I guess new translation repositories can just copy/paste and adjust the our GitHub Actions setup from us, and have it working for them quite easily. So maybe we do not need to bother @jnavila anymore? 😉

### DIFF
--- a/TRANSLATING.md
+++ b/TRANSLATING.md
@@ -82,7 +82,7 @@ GitHub assumes that forks will be used to contribute to the repo from which they
 
 ## Setting up a publication chain for e-books
 
-This is a technical task, please ping @jnavila to get started with epub publication.
+This is a technical task, please ping [@jnavila](https://github.com/jnavila) to get started with epub publication.
 
 ## Beyond Pro Git
 


ben:
> I guess new translation repositories can just copy/paste and adjust the our GitHub Actions setup from us

Is that enough, though? IIRC we need to do something on the git-scm.com website to list the new files for download, right?

HonkingGoose:
> Is that enough, though? IIRC we need to do something on the git-scm.com website to list the new files for download, right?

@ben You're correct, you need to add the new translation to a file on the `git/git-scm.com` repository. Quote from the [ARCHITECTURE.md file on `git/git-scm.com`](https://github.com/git/git-scm.com/blob/main/ARCHITECTURE.md#manual-intervention):

> ## Manual Intervention
> 
> The site mostly just runs without intervention:
> 
>   - code merged to `main` is auto-deployed
> 
>   - new git versions are detected daily and manpages and download links
>     updated
> 
>   - book updates (including translations) are picked up daily
> 
> There are a few tasks that still need to be handled by a human:
> 
>   - new images added to the book have to be copied manually from
>     progit/progit2
> 
>   - new languages for book translations need to be added to
>     `lib/tasks/book2.rake`
> 
>   - forced re-imports of content (e.g., a formatting fix to imported
>     manpages) must be triggered manually

jnavila:
Hello,

Translators can do this all by themselves, but they usually don't have the knowledge to do so. As I have to manage a lot of repositories the same way, I have set up a repo with the shared recipes for producing the ebooks https://github.com/progit/progit2-pub where all the translations repos pull the build scripts. These scripts add two things compared to progit:
 * internationalization of the compilation. Asciidoctor needs some additional direction when compiling some languages
 * health check up of the book by using epubcheck. This is a very crude way of checking that the structure of the book is still valid but it helps a lot the contributors to deal with asciidoc's syntax.

This used to work well until travis-ci.org shut down. I haven't yet switched to github actions, by lack of time. The idea is to put minimal stuff in the CI config. The CI part still needs to provide the tagging and publishing of artifacts. I will copy @ben 's part for this. I intend to use a [GitHub action template](https://docs.github.com/en/actions/learn-github-actions/sharing-workflows-with-your-organization) for this.

If it is of interest, we could switch progit itself to this workflow (the one using progit2-pub), so that when translators would fork it, they would already have the CI and publishing scripts ready for git-scm.com. What do you think?

HonkingGoose:
I like the idea of making a sort of template that others can then use with minimal work! I think using GitHub Actions is easier than Travis CI, as the hosting is on GitHub anyways.

I also like the idea of standardizing the setup accross the organization. This makes it so its easy to debug, because we all use the same stuff.

I'll let @ben chime in with his thoughts. 😄

HonkingGoose:
I've made a new issue where we can track/discuss the idea to "standardize our setup across the organization in some way", see #1691 for that. 😉 

Do we want the change in this PR? Or would we rather leave the text as-is?

jnavila:
I thought that using @ with username was enough to bring up a notification to the designated user.

HonkingGoose:
> I thought that using @ with username was enough to bring up a notification to the designated user.

That only works for issues/pull requests/discussions, so things coming from within GitHub somehow.

Right now we use the `@ with username` pattern, and that does not turn into a link when you view the file on GitHub: https://github.com/progit/progit2/blob/main/TRANSLATING.md#setting-up-a-publication-chain-for-e-books

That's why I want to add an explicit link to your profile on GitHub. 😉

PR #1689, opened 2021-07-20
Point readers to Libera Chat IRC

knittl:
The #git channel on Freenode does not have an active user base anymore.
Match the recommended IRC network of <https://git-scm.com/community>.

## Changes

-


ben:
👍🏼 Thanks!

PR #1692, opened 2021-07-26
Create  TOSECURITY.md

Tomtam25:

## Changes

-

### DIFF
--- a/TOSECURITY.md
+++ b/TOSECURITY.md
@@ -1,0 +1,21 @@
+# Security Policy
+
+## Supported Versions
+
+Use this section to tell people about which versions of your project are
+currently being supported with security updates.
+
+| Version | Supported          |
+| ------- | ------------------ |
+| 5.1.x   | :white_check_mark: |
+| 5.0.x   | :x:                |
+| 4.0.x   | :white_check_mark: |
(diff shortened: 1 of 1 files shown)


ben:
I don't think this really applies to us? Can you give me more of an idea of what you're trying to do here?

HonkingGoose:
> I don't think this really applies to us? Can you give me more of an idea of what you're trying to do here?

This is just the basic GitHub security template, that you get when you go to the Security -> Overview -> Security policy section of this repository. 😉 

Normally I'd be called `SECURITY.md` and you'd put information in there on how people should report security bugs/concerns. Sometimes you _really_ want people to email instead of opening an issue. This can give the repository maintainer time to respond/fix the problem before all the world knows about it.

Example of a real world `SECURITY.md` file: https://github.com/renovatebot/renovate/blob/main/SECURITY.md

ben:
Okay, now that I'm thinking about it, we could have a channel for disclosing some kind of security issue. I'm having a hard time imagining one, because none of our outputs include running code, but unknown-unknowns, etc. Fine.

This PR as of today, nearly a month into its life, is still just template content. @Tomtam25 do you intend to fill this in? Is it your intent to get _us_ to fill it in?

PR #1698, opened 2021-08-09
Fix quotes in "Rewriting history" section

matshch:

## Changes
- Quotation marks style is changed from grave+apostrophe to combination that is properly rendered as slanted double quotation marks as used in other parts of the book.

## Context
This is the only place in the book where this type of quotes were used. Also, grave in the Noto Serif font included with EPUB version of the book is combining, so this two words were rendered incorrectly:
Screenshot_20210809-201914__01


ben:
✨ Thanks!

PR #1699, opened 2021-08-09
Use proper required checkboxes syntax

HonkingGoose:

## Changes

- Use proper required checkboxes syntax

## Context

The old forms are not properly forcing some checkboxes to be ticked before proceeding, this PR makes it so that we properly force the checkboxes. 😄 

Turns out you need to put the `required: true` value under the `option` key, not in a separate `validations` key... 😄 

Example of _invalid_ syntax:

```yaml
  - type: checkboxes
    attributes:
      label: There's no existing/similar bug report.
      description: Please search to see if an issue already exists for the bug you encountered.
      options:
        - label: I have searched the existing issues
    validations:
      required: true
```

Example of _correct_ syntax:

```yaml
  - type: checkboxes
    attributes:
      label: There's no existing/similar bug report.
      description: Please search to see if an issue already exists for the bug you encountered.
      options:
        - label: I have searched the existing issues
          required: true
```

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -8,8 +8,7 @@ body:
       description: Please search to see if an issue already exists for the bug you encountered.
       options:
         - label: I have searched the existing issues
-    validations:
-      required: true
+          required: true
 
   - type: checkboxes
     attributes:
@@ -17,8 +16,7 @@ body:
       description: Please create one issue per bug, split up your bug reports if needed.
       options:
         - label: I'm reporting a single actionable bug
-    validations:
-      required: true
+          required: true
 
   - type: checkboxes
     attributes:
@@ -26,8 +24,7 @@ body:
       description: Bug reports about translations or the first version of the book are not accepted.
       options:
         - label: This bug is not about a translation or old version
-    validations:
-      required: true
+          required: true
 
   - type: checkboxes
     attributes:


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -8,26 +8,23 @@ body:
...
     attributes:
       label: This report is about the ProGit book, version 2, English language.
       description: Bug reports about translations or the first version of the book are not accepted.
       options:
         - label: This bug is not about a translation or old version
-    validations:
-      required: true
+          required: true
 
   - type: checkboxes
```
HonkingGoose, line 29: Do we _require_ the following checkboxes to be ticked before proceeding?

- [ ] I confirm the bug is about the book as found on the website/pdf
- [ ] This bug also affects the Pro Git book as published on the website

Let me know if I need to require one or both of these checkboxes as well, and I'll include those changes in this PR.
That way we fix all checkboxes at once. 😄

ben: Nah, let them be optional for now. Ready to merge?

HonkingGoose: Let's keep those optional then.

HonkingGoose: @ben We're good to merge!

Terrybrar97 (approved):
Ok

PR #1703, opened 2021-08-25
Create SECURITY.md

HonkingGoose:

## Changes

- Create a proper functional `SECURITY.md` file

## Context

I'm taking the liberty to create a proper PR, as PR #1692 is bad, and should not be merged as-is.

The idea here is that we redirect any security concerns to a private email address, instead of people opening a issue on the repository, or discussing their concerns in public in some other way.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,5 @@
+## Reporting a security issue
+
+If you find any security issue or vulnerability, please email us at INSERT EMAIL HERE with your report.
+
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -0,0 +1,5 @@
+## Reporting a security issue
+
+If you find any security issue or vulnerability, please email us at INSERT EMAIL HERE with your report.
```
HonkingGoose, line 3: @ben Please make a suggestion for the correct email adress, as I don't know which email you want to use. 😉

PR #1715, opened 2021-11-22
Update message to be equal with later mentions

jo3rn:

## Changes

commit message in `maintenance.asc` (chapter 10)
"Modify repo.rb a bit" gets used later on in the log outputs for that same commit, so the message should be equal.


ben:
Nice catch, thanks!

PR #1716, opened 2021-11-29
update link

benkoshy:
* allows readers to follow along by typing in the commands.

(Note: running `GET info/refs` after cloning the repository yeilds unexpected results for me: the command line dumps html which seems to want for me to prove i'm a human via a recaptcha test? Not sure what to do about that)

### DIFF
--- a/book/10-git-internals/sections/transfer-protocols.asc
+++ b/book/10-git-internals/sections/transfer-protocols.asc
@@ -19,7 +19,7 @@ Let's follow the `http-fetch` process for the simplegit library:
 
 [source,console]
 ----
-$ git clone http://server/simplegit-progit.git
+$ git clone https://github.com/schacon/simplegit-progit.git
 ----
 
 The first thing this command does is pull down the `info/refs` file.


ben:
What this section is describing isn't of any practical value at this point in history. GitHub [turned off](https://github.blog/2011-03-09-git-dumb-http-transport-to-be-turned-off-in-90-days/) the insecure dumb-http protocol 10 years ago, so you'll have to stand up your own Git HTTP server to be able to test any of this yourself. The section is about how clone works in a conceptual way, not in a manually-reproducible way.

Given that, I'm of two minds about this patch. On the one hand, giving a real repo people can visit to see what the commands are talking about is nice. On the other hand, it sends a mixed signal that you should be able to follow along in a shell, when you actually can't. What do you think?

benkoshy:
> On the other hand, it sends a mixed signal that you should be able to follow along in a shell, when you actually can't. What do you think?

Agreed - following along / experimenting in the shell is a nice learning experience. because I was attempting to follow along manually (e.g. typing in commands) -- and given nothing worked  (the link was wrong? and when i found the link it did not work - presumably because, as stated, github turned off that feature 10 years ago), I was very confused - the train of thought might have gone something like this: "am i doing something wrong? i'm a newbie learning git and something doesn't seem right. shrugs shoulders" 

> What do you think?

To answer your question, I can enumerate a few ideas: 

- we could clearly state that these commands don't work without a non-smart server
- we provide a non-smart http server (i.e. a sandbox) - perhaps a separate docker / vagrant sandbox where these experiments can be conducted (not sure how feasible this would be -- just thinking out loud).
- we could rewrite the section to focus on the[ concepts with diagrams / lucid explanations](https://stackoverflow.com/a/51646150/4880924) vs the current approach (that of typing out commands).

Since this is your project/creation ultimately it's your call - I figured I would merely present my ideas as a reader. (btw i thank you for work efforts on the book etc).

i would be curious to know your thoughts, and i would not be averse to further PRs / ideas if required (depending on my capacity).

revmarcuspeter-cyber (approved):

PR #1729, opened 2022-02-06
Note that master is not always the default

noureddin:

## Changes

- Add an id to Ch01 > First-time Git Setup > Your default branch name.
- Add a note in Ch02 > Recording Changes to the Repository > Checking the Status of Your Files, directly after the paragraph that says:

  > Finally, the command tells you which branch you’re on and informs you that it has not diverged from the same branch on the server. For now, that branch is always `master`, which is the default; you won’t worry about it here.

  The note points out that remote repositories are likely to have a different default branch name.

## Context

GitHub in mid-2020 changed the default branch name from `master` to `main`, and so did other Git hosts.
And many developers use an even different default branch name like `dev` or `develop` or something else.

A beginner might be very confused when seeing a repo whose default branch is not `master`. So I think it worth pointing out that explicitly.

I mean, the book asserts that the default branch is always `master`, when explaining the `git status` output in a freshly cloned repo. That repo is [libgit2](https://github.com/libgit2/libgit2), whose default branch is `main`…

### DIFF
--- a/book/01-introduction/sections/first-time-setup.asc
+++ b/book/01-introduction/sections/first-time-setup.asc
@@ -84,6 +84,7 @@ You may find, if you don't setup your editor like this, you get into a really co
 An example on a Windows system may include a prematurely terminated Git operation during a Git initiated edit.
 ====
 
+[[_default_branch]]
 ==== Your default branch name
 
 By default Git will create a branch called _master_ when you create a new repository with `git init`.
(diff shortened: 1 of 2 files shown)


HonkingGoose:
I found a related issue: #1462

GitHub and GitHub now create a `main` branch for you when you create a new repository from their interface.
I don't know when the Git program will make `main` the default branch when you run `git init`.

I guess when the book was written the `libgit2` repo used `master`, and they migrated to `main` some time after the book was written. 😉

ben (approved):
✨

PR #1738, opened 2022-05-03
ch-10: avoid unnecessary mention update-index's --add

sivaraam:

## Changes

The book clearly mentions that --add needs to be passed to 'git update-index'
only when the file is new to the index. When the second version of 'test.txt' is
added, the index know about the file already.

So, avoid passing the '--add' when adding the second version of 'test.txt'


ben:
Nice. Thanks!

PR #1763, opened 2022-06-15
Fix user.signingkey example

kariya-mitsuru:

## Changes

Since the value of user.signingkey is passed unchanged to gpg's --local-user parameter, the fingerprint needs to be suffixed with an exclamation mark (!).
If a fingerprint is specified without an exclamation mark, it is simply ignored.


ben:
Awesome. Thanks!

PR #1774, opened 2022-07-16
Remove unnecessary dig at MS Word

danielhollas:

## Changes
I just removed one sentence that "comments" on MS Word. Either everybody agrees with the sentiment, making the sentence redundant, or not, making it alienating to some. Author's view is already apparent from the previous sentence. :smiley: 

btw: Thanks so much for a great resource! :clap:


ben:
Yeah, that's fair. 😊 Thanks!

PR #1776, opened 2022-07-17
Mark up "input" as in-line code

solt87:

## Changes

- Adds in-line code markup to the word "input",
  as it is mentioned as a setting value.

## Context

When mentioning the `core.autocrlf` value `true`, the text marks it up as in-line code.
This patch applies the same markup to the `input` value of the same setting.


ben:
✨

PR #1789, opened 2022-08-29
Use caret shorthand to open links in new tab

### LINKED ISSUE #1604: Open external links in new browser window
PR #1603 fixed a mistaken Markdown link introduced by me.

I noticed the PR creator uses a caret `^` behind the link.
This makes it so that the link is opened in a new browser window (`_blank`).

I think it makes sense to open external links (pages not part of book) in a new tab in the browser.
This way users don't accidentally navigate away from the book.

Quote from [Asciidoctor writers guide, links and images](https://asciidoctor.org/docs/asciidoc-writers-guide/#links-and-images):

> Since `_blank` is the most common window name, we’ve introduced shorthand for it. Just end the link text with a caret (`^`):

```
Let's view the raw HTML of the link:view-source:asciidoctor.org[Asciidoctor homepage^].
```

HonkingGoose:

## Changes

- Use the caret `^` shorthand to open links in a new tab [^docs]
- Drop a double newline at the end of a file

## Context

Fixes #1604.

[^docs]: https://asciidoctor.org/docs/asciidoc-writers-guide/#target-window-and-role-attributes-for-links

### DIFF
--- a/book/01-introduction/sections/about-version-control.asc
+++ b/book/01-introduction/sections/about-version-control.asc
@@ -23,7 +23,7 @@ To deal with this issue, programmers long ago developed local VCSs that had a si
 image::images/local.png[Local version control diagram]
 
 One of the most popular VCS tools was a system called RCS, which is still distributed with many computers today.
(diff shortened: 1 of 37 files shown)


ben (approved):
✨

PR #1797, opened 2022-09-27
Update snapshots.svg

chrisdecker1201:
I had an issue using this graphic in a paper together with Latex and the SVG package. Because somehow the Version1-5 text wasn't white like the deltas.svg. It seems, that the style="" did not work there. I looked into the deltas.svg file and updated this. Maybe it's useful for you, too.


ben:
:sparkles: Thanks!

PR #1799, opened 2022-10-03
Update what-is-git.asc

aziz58:

## Changes

-

### DIFF
--- a/book/01-introduction/sections/what-is-git.asc
+++ b/book/01-introduction/sections/what-is-git.asc
@@ -83,7 +83,7 @@ Git has three main states that your files can reside in: _modified_, _staged_, a
 * Staged means that you have marked a modified file in its current version to go into your next commit snapshot.
 * Committed means that the data is safely stored in your local database.
 
-This leads us to the three main sections of a Git project: the working tree, the staging area, and the Git directory.
+This leads us to the three main sections of a Git project: the working directory, the staging area, and the Git directory.
 
 .Working tree, staging area, and Git directory
 image::images/areas.png["Working tree, staging area, and Git directory."]


book/01-introduction/sections/what-is-git.asc
```diff
@@ -83,7 +83,7 @@ Git has three main states that your files can reside in: _modified_, _staged_, a
 * Staged means that you have marked a modified file in its current version to go into your next commit snapshot.
 * Committed means that the data is safely stored in your local database.
 
-This leads us to the three main sections of a Git project: the working tree, the staging area, and the Git directory.
+This leads us to the three main sections of a Git project: the working directory, the staging area, and the Git directory.
 
 .Working tree, staging area, and Git directory
 image::images/areas.png["Working tree, staging area, and Git directory."]
```
ben, line 89: Unfortunately this change disagrees with the caption of the diagram right below it. The diagram itself says "Working Directory", so let's standardize on that. You'll want to update all of these locations.

CleanShot 2022-10-05 at 11 00 38

solt87:
I assume the intent of this change is to harmonize the text with the diagram and its caption, which to me seems self-evident. But I think there's a bigger issue here: the portion on the screenshot seems to assume that "working tree" is the same as "working directory", and this change would use "working directory" consistently (at least taking into account @ben's suggestion). However, according to the Git [glossary](https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-aiddeftreeatree), "tree", "tree object" and "working tree" mean different things, with "tree" being _"[...] equivalent to a [directory](https://git-scm.com/docs/gitglossary#def_directory)"_. Whereas the "working tree" _"[...] normally contains the contents of the HEAD commit’s tree, **plus any local changes** that you have made but not yet committed."_ (My emphasis.)

I think the two decent options would be to either explain both "working tree" and "working directory", and use the terms consistently and correctly, or (and maybe this is preferable in an introductory textbook) use the two terms interchangeably, but mention in parentheses that they aren't exactly the same, with a reference to the glossary.

(I won't even bring up the fact that there is another related term, "worktree", which seems to be different from "working tree", as this complicates an already convoluted matter further. :) )

PR #1817, opened 2022-12-18
Add FB2 output

slonopotamus:

## Changes

- Add generation of FB2 document


ben:
OK, sure, why not. Thanks!

PR #1843, opened 2023-03-15
Use caret shorthand and tags URL

petk:

## Changes

- This adds two caret shorthands to the URL in the installation and contributing chapters so they open in new tabs
- The GitHub mirror URL for Git repository points to empty releases page. Probably the URL with tags might be better.


ben (approved):
✨ Thanks!

PR #1844, opened 2023-03-16
Correction recording-changes.asc

tejasps2001:

Modified the statement regarding the **'git diff'** command to clarify that it compares the changes between the files in your working directory and the files in your Git repository, and it shows the differences between the two sources. By default, it displays the changes that are not yet staged for commit. However, if you use the **'--staged'** or **'--cached'** option, **'git diff'** will compare the changes between the files in your staging area and the latest commit in your Git repository.

### DIFF
--- a/book/02-git-basics/sections/recording-changes.asc
+++ b/book/02-git-basics/sections/recording-changes.asc
@@ -325,7 +325,7 @@ index 8ebb991..643e24f 100644
  that highlights your work in progress (and note in the PR title that it's
 ----
 
-That command compares what is in your working directory with what is in your staging area.
+That command compares what is in your working directory with what is in your Git repository.
 The result tells you the changes you've made that you haven't yet staged.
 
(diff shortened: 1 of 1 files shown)


ben:
Going to disagree with your change here. Here's a paraphrased version of what happens in [that section](https://git-scm.com/book/en/v2/Git-Basics-Recording-Changes-to-the-Repository#_git_diff_staged):

- Edit the `README` and the `CONTRIBUTING.md` files
- `git add README`
- `git diff`
- See that only the changes to `CONTRIBUTING.md` are shown
- `git diff --staged`
- See that only the changes to `README` are shown

So we _do_ cover the `--staged` flag just after the section you're changing. What's more, while your change sounds friendlier, it's inventing this new concept of things being "in the repo." Is the working directory not in the repo? I'd argue that it is.

tejasps2001:
Thank you for your response and for taking the time to consider my suggestion.

PR #1847, opened 2023-03-25
Use lowercase

petk:

## Changes

- Sentences inside brackets are lowercased elsewhere in the book, so this syncs the same capitalization also on two remaining places.


ben (approved):
👍🏼

PR #1866, opened 2023-04-14
Fix few punctuation marks

petk:

## Changes

- This changes few punctuation marks (if I'm not mistaken, they should be outside of the formatted text).


petk:
Ah, I've rechecked this one and it is correct to put the punctuation marks in the quotes. Well, I guess we learn every day something new. :D

PR #1881, opened 2023-05-12
Fix the next version calculation

petk:

## Changes

- Version written in the book was one tag behind the one on the PDF at GitHub releases. Next tag needed to be calculated here.

Steps to reproduce the issue:

- Visit https://git-scm.com/book/en/v2
- Download PDF
- Version in the PDF and the last published tag on the GitHub releases don't match (one patch behind)

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,9 +1,12 @@
 namespace :book do
 
   # Variables referenced for build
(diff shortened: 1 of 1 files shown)


ben:
NIce, thanks!

PR #1926, opened 2024-02-02
fixed typo in chapter 8 under core.pager

### LINKED ISSUE #1925: If you disable the pager, Git will still page the output?
Under `core.pager`, it talks about disabling the pager, but then it says Git will still page the output. Is that a typo? Should it say "Git will **print**" instead? Or maybe it's not clear what the word "page" means.

https://github.com/progit/progit2/blob/72e19595310d0e3b45b6257692c5297800fae06b/book/08-customizing-git/sections/config.asc?plain=1#L128-L132

For context, the [full section](https://github.com/progit/progit2/blob/72e19595310d0e3b45b6257692c5297800fae06b/book/08-customizing-git/sections/config.asc#corepager):

<blockquote>

##### `core.pager`

This setting determines which pager is used when Git pages output such as `log` and `diff`. You can set it to `more` or to your favorite pager (by default, it's `less`), or you can turn it off by setting it to a blank string:

```text-shell-session
$ git config --global core.pager ''
```

If you run that, Git will page the entire output of all commands, no matter how long they are.

</blockquote>

shyena61:
There is a typo in chapter 8 under core.pager last line, it say 'Git will page'. The word 'page' should be replaced with the word 'print'

## Changes

- Fixed typo in last line of chapter 8 under core.pager.
- The word "page" is replaced by "print".

## Context

Fixes #1925

### DIFF
--- a/book/08-customizing-git/sections/config.asc
+++ b/book/08-customizing-git/sections/config.asc
@@ -129,7 +129,7 @@ You can set it to `more` or to your favorite pager (by default, it's `less`), or
 $ git config --global core.pager ''
 ----
 
-If you run that, Git will page the entire output of all commands, no matter how long they are.
+If you run that, Git will print the entire output of all commands, no matter how long they are.
 
 ===== `user.signingkey`
 


ben (approved):
Nice, thanks! ✨

PR #1930, opened 2024-02-12
specify which Update>> prompt is referred to in interactive-staging

Mmmmmmichael:

## Context
A somewhat inattentive reader (me, when I read the chapter) may be temporarily confused about which of the `Update>>`prompts the (in this PR edited) sentence is referring to as the previous prompt was also an `Update>>` prompt.
I believe this change would make the chapter slightly more easily understandable, but if you disagree and think I am the only person this would ever happen to (which is possible, the PR is merely based on my personal reading experience) or that this is not worth changing, then you can safely ignore and close the PR.

The following is the excerpt this change is about:
```asc
To stage the `TODO` and `index.html` files, you can type the numbers:

[source,console]
----
Update>> 1,2
           staged     unstaged path
* 1:    unchanged        +0/-1 TODO
* 2:    unchanged        +1/-1 index.html
  3:    unchanged        +5/-1 lib/simplegit.rb
Update>>
----

The `*` next to each file means the file is selected to be staged.
If you press Enter after typing nothing at the `Update>>` prompt, Git takes anything selected and stages it for you:
```

### DIFF
--- a/book/07-git-tools/sections/interactive-staging.asc
+++ b/book/07-git-tools/sections/interactive-staging.asc
@@ -53,7 +53,7 @@ Update>>
 ----
 
 The `*` next to each file means the file is selected to be staged.
-If you press Enter after typing nothing at the `Update>>` prompt, Git takes anything selected and stages it for you:
+If you press Enter after typing nothing at the second `Update>>` prompt, Git takes anything selected and stages it for you:
 
 [source,console]
 ----


michaelblyons (approved):
Looks good to me.

revmarcuspeter-cyber (approved):

revmarcuspeter-cyber (approved):

PR #1932, opened 2024-02-20
Simple clarification

hamid-nazari:
- [ x ] I provide my work under the [project license](https://github.com/progit/progit2/blob/main/LICENSE.asc).
- [ x ] I grant such license of my work as is required for the purposes of future print editions to [Ben Straub](https://github.com/ben) and [Scott Chacon](https://github.com/schacon).

## Changes

I think `and then` would be better changed to `and when later` or `and when (later)` to better clarify the purpose (I got confused the first time).

> ... 
Then you need to tell Git what to do if you have made changes and **when later** `git submodule update --remote` pulls in new work from upstream.

## Context

My change is on `Git Tools/Submodules/Working on a Project with Submodules/Working on a Submodule`

### DIFF
--- a/book/07-git-tools/sections/submodules.asc
+++ b/book/07-git-tools/sections/submodules.asc
@@ -466,7 +466,7 @@ You have to do some extra steps if you want changes in a submodule to be tracked
 
 In order to set up your submodule to be easier to go in and hack on, you need to do two things.
 You need to go into each submodule and check out a branch to work on.
-Then you need to tell Git what to do if you have made changes and when later `git submodule update --remote` pulls in new work from upstream.
+Then you need to tell Git what to do if you have made changes and later `git submodule update --remote` pulls in new work from upstream.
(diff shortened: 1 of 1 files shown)


book/07-git-tools/sections/submodules.asc
```diff
@@ -466,7 +466,7 @@ You have to do some extra steps if you want changes in a submodule to be tracked
 
 In order to set up your submodule to be easier to go in and hack on, you need to do two things.
 You need to go into each submodule and check out a branch to work on.
-Then you need to tell Git what to do if you have made changes and then `git submodule update --remote` pulls in new work from upstream.
+Then you need to tell Git what to do if you have made changes and when later `git submodule update --remote` pulls in new work from upstream.
```
ben, line 469: How about this?

```suggestion
Then you need to tell Git what to do if you have made changes and later `git submodule update --remote` pulls in new work from upstream.
```

hamid-nazari: That is also fine and to the point

PR #1970, opened 2024-08-09
Escape notepad++ default installation path

z-hed:

## Changes

- Add a single backslash to the Notepad++ editor path in A3.1 Setup and Config.

## Context

It appears that the pair of "++" in the Notepad++ path may be interpreted as an inline passthrough by Asciidoctor. The double pluses are omitted in the rendered text (both on git-scm.com and the .pdf in the latest release.) The resulting command is incorrect if copied and pasted into the terminal, which is exactly what I had done when I noticed the problem.

If GitHub's web editor is rendering my change correctly, a single backslash in the first pair of double pluses is enough to fix the issue. After reading the contributing guidelines, I figured this is a minor enough correction that I did not need to open an issue first, but I have not contributed before. Constructive criticism welcomed. Thank you for your time.

### DIFF
--- a/C-git-commands.asc
+++ b/C-git-commands.asc
@@ -59,7 +59,7 @@ Accompanying the configuration instructions in <<ch01-getting-started#_editor>>,
 |Kate (Linux) |`git config --global core.editor "kate --block"`
 |nano |`git config --global core.editor "nano -w"`
 |Notepad (Windows 64-bit) |`git config core.editor notepad`
-|Notepad++ (Windows 64-bit) |`git config --global core.editor "'C:\Program Files\Notepad++\notepad++.exe' -multiInst -notabbar -nosession -noPlugin"` (Also see note below)
+|Notepad++ (Windows 64-bit) |`git config --global core.editor "'C:\Program Files\Notepad+\+\notepad++.exe' -multiInst -notabbar -nosession -noPlugin"` (Also see note below)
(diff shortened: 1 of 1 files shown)


C-git-commands.asc
```diff
@@ -59,7 +59,7 @@ Accompanying the configuration instructions in <<ch01-getting-started#_editor>>,
 |Kate (Linux) |`git config --global core.editor "kate --block"`
 |nano |`git config --global core.editor "nano -w"`
 |Notepad (Windows 64-bit) |`git config core.editor notepad`
-|Notepad++ (Windows 64-bit) |`git config --global core.editor "'C:\Program Files\Notepad++\notepad++.exe' -multiInst -notabbar -nosession -noPlugin"` (Also see note below)
+|Notepad++ (Windows 64-bit) |`git config --global core.editor "'C:\Program Files\Notepad+\+\notepad++.exe' -multiInst -notabbar -nosession -noPlugin"` (Also see note below)
```
ben, line 62: Interesting. Only the second `+` needs to be escaped? The other three are fine?

z-hed: I believe escaping any one of the four `+` that surrounds `\notepad` should stop it from being considered as an enclosing pair of double pluses.

When I experimented with escaping more than one, the literal backslashes appeared in the output, possibly because by that point there was no special meaning to actually escape and they were interpreted literally.

ben (approved):

PR #1973, opened 2024-08-12
Update viewing-history.asc

shallow-beach:

## Changes

- make explicit note about shortcut

## Context

Draws attention to a shortcut not previously demonstrated. Easily missed (by me).


michaelblyons:
I think instead of adding a note there, it might be better to modify the text in the table on line 223.

``` diff
-| `--pretty`        | Show commits in an alternate format. Option values include `oneline`, `short`, `full`, `fuller`, and `format` (where you specify your own format).
+| `--pretty`        | Show commits in an alternate format. Option values include `oneline`, `short`, `full`, `fuller`, or specify your own format.
```

revmarcuspeter-cyber (approved):

PR #1982, opened 2024-09-13
Fixed: error in topic-branches-2.svg

iidear:

## Changes

- **Chapter 3.4** *Figure 29. History after merging dumbidea and iss91v2.* C14 should not be ahead of C11
- topic-branches-2.png needs modification too, but I don't know how to generate the same quality png


rosshjb:
I don't think so:
> C14 should not be ahead of C11

IMO, there is nothing wrong with the original image. The figure 29 shows that the `dumbidea` and `issq1v2` branches have been merged into the `master` branch. Following the diagram you suggested, the `issq1v2` branch hasn't merged into the `master` branch. Refer the following paragraph:

> Now, let’s say you decide you like the second solution to your issue best (`iss91v2`); and you showed the `dumbidea` branch to your coworkers, and it turns out to be genius. You can throw away the original `iss91` branch (losing commits `C5` and `C6`) and merge in the other two.

PR #2021, opened 2025-02-15
Remove recommendation for git fetch && git merge over git pull

### LINKED ISSUE #2020: Poorly justified usage recommendation about `git pull` vs `git fetch`
### General overview of your idea.

Chapter 3, section "Remote Branches" contains the following line:
"Generally it’s better to simply use the `fetch` and `merge` commands explicitly as the magic of git pull can often be confusing."

This seems out of touch with reality. I don't think I know a single person who does this: it's an unnecessary extra step and `git pull` isn't that complicated. Using it directly is usually not a problem.
In my opinion, this line should just be removed. 

### What problem will this solve?

The current text misleads users into suboptimal behavior, and removing it would solve that problem, or at least let them come to their own conclusions based on the facts presented in the chapter.

### Have you thought about other solutions?

As an alternative, the section should further elaborate explicitly why `git pull` is not recommended and what problems it can cause. In any case, the user should be allowed to draw their own conclusion.

### Do you want to help with this enhancement idea?

Yes

DeinAlptraum:

## Changes
Remove the recommendation for `git fetch && git merge` over `git pull`

## Context
Fixes #2020


LTBD-Admin (commented):
Using git fetch followed by git merge gives developers more control over the merging process. It allows them to review changes before merging, which can help avoid unexpected conflicts.

PR #2023, opened 2025-03-10
Clarify that --global configuration is not applicable system-wide

trevorjobling:
Clarify that --global configuration is not applicable system-wide, but only within the scope of the logged in as current user.

## Changes
add "while logged in as you"

## Context
Configuration via the  `--global` option does not affect everything "on that system". I believe it is limited in scope to the user. This change attempt to make that clear. It could be worded better but I'm struggling to come with anything better than this!

### DIFF
--- a/book/01-introduction/sections/first-time-setup.asc
+++ b/book/01-introduction/sections/first-time-setup.asc
@@ -42,7 +42,7 @@ $ git config --global user.name "John Doe"
 $ git config --global user.email johndoe@example.com
 ----
 
-Again, you need to do this only once if you pass the `--global` option, because then Git will always use that information for anything you do while logged in as you on that system.
+Again, you need to do this only once if you pass the `--global` option, because then Git will always use that information for your user on that system.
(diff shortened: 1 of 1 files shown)


book/01-introduction/sections/first-time-setup.asc
```diff
@@ -42,7 +42,7 @@ $ git config --global user.name "John Doe"
 $ git config --global user.email johndoe@example.com
 ----
 
-Again, you need to do this only once if you pass the `--global` option, because then Git will always use that information for anything you do on that system.
+Again, you need to do this only once if you pass the `--global` option, because then Git will always use that information for anything you do while logged in as you on that system.
```
michaelblyons, line 45:
```suggestion
Again, you need to do this only once if you pass the `--global` option, because then Git will always use that information for your user on that system.
```

trevorjobling: Shorter _and_ clearer. Thanks!

ben (approved):
✨ Thanks!

PR #2065, opened 2026-01-09
Fix command substitution syntax in archive examples

JiaqiFan42:
Make the command subsititution syntax modern and consistent. (e.g. the example on line 282 already used the "$(...)" syntax. I suppose it's better to make it consistent.)

## Changes

- Make the command subsititution syntax modern and consistent. (e.g. the example on line 282 already used the "$(...)" syntax. I suppose it's better to make it consistent.)


gigako1981:
Owner ``

